PR #42, opened 2015-11-21
Capture Jenkins job names correctly when the folder plugin is being used

AMeng:
This is meant to address https://github.com/spinnaker/spinnaker/issues/605.
I'm able to hit `http://localhost:8080/jobs/<JenkinsName>/` and see a list of jobs with the expected names:

```json
[
  "Job1",
  "Folder1/job/Job2",
  "Folder1/job/Folder2/job/Job3"
]
```
Its also likely that a few endpoints will need to be updated since they currently have the job names in the middle of other parameters, such as "/jobs/{jobName}/builds" which could be troublesome.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/InfoController.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/InfoController.groovy
@@ -22,6 +22,7 @@ import com.netflix.spinnaker.igor.jenkins.client.model.Build
 import com.netflix.spinnaker.igor.jenkins.client.model.JobConfig
 import groovy.transform.InheritConstructors
 import groovy.util.logging.Slf4j
+import javax.servlet.http.HttpServletRequest
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.http.HttpStatus
 import org.springframework.web.bind.annotation.PathVariable
@@ -30,6 +31,7 @@ import org.springframework.web.bind.annotation.RequestMethod
 import org.springframework.web.bind.annotation.RequestParam
 import org.springframework.web.bind.annotation.ResponseStatus
 import org.springframework.web.bind.annotation.RestController
+import org.springframework.web.servlet.HandlerMapping
 
 import javax.ws.rs.QueryParam
 
@@ -73,11 +75,31 @@ class InfoController {
             throw new MasterNotFoundException("Master '${master}' does not exist")
         }
 
-        jenkinsService.jobs.list.collect { it.name }
+        def jobList = []
+        def recursiveGetJobs
+
+        recursiveGetJobs = { list, prefix="" ->
+            if (prefix) {
+                prefix = prefix + "/job/"
+            }
+            list.each {
+                if (it.list == null || it.list.empty) {
+                    jobList << prefix + it.name
+                } else {
+                    recursiveGetJobs(it.list, prefix + it.name)
+                }
+            }
+        }
+        recursiveGetJobs(jenkinsService.jobs.list)
+
+        return jobList
     }
 
-    @RequestMapping(value = '/jobs/{master}/{job:.+}')
-    JobConfig getJobConfig(@PathVariable String master, @PathVariable String job) {
+    @RequestMapping(value = '/jobs/{master}/**')
+    JobConfig getJobConfig(@PathVariable String master, HttpServletRequest request) {
+        def job = (String) request.getAttribute(
+            HandlerMapping.PATH_WITHIN_HANDLER_MAPPING_ATTRIBUTE).split('/').drop(3).join('/')
+
         log.info('Getting the job config for {} at {}', job, master)
 
         def jenkinsService = masters.map[master]
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/client/JenkinsClient.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/client/JenkinsClient.groovy
@@ -25,6 +25,7 @@ import com.netflix.spinnaker.igor.jenkins.client.model.ProjectsList
 import com.netflix.spinnaker.igor.jenkins.client.model.QueuedJob
 import com.netflix.spinnaker.igor.jenkins.client.model.ScmDetails
 import retrofit.client.Response
+import retrofit.http.EncodedPath
 import retrofit.http.GET
 import retrofit.http.POST
 import retrofit.http.Path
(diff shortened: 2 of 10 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/InfoController.groovy
```diff
@@ -73,7 +73,24 @@ class InfoController {
             throw new MasterNotFoundException("Master '${master}' does not exist")
         }
 
-        jenkinsClient.jobs.list.collect { it.name }
+        def jobList = []
```
tomaslin: would like to see a corresponding test in https://github.com/spinnaker/igor/blob/master/igor-web/src/test/groovy/com/netflix/spinnaker/igor/jenkins/InfoControllerSpec.groovy

tomaslin:
This is failing for me when I am just trying to get the job details after installing the jenkins folders plugin on a jenkins host. There doesn't seem to be a straightforward way to modify spring paths to accept / 

https://github.com/spinnaker/igor/blob/master/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/InfoController.groovy#L79 

I have this job

```
[
"Jenkins folder/job/hakuna matata"
]
```

under the host jenkins1. When I try to get the details for it, it doesn't appear to handle the path correctly:

http://localhost:8080/jobs/jenkins1/Jenkins%20folder/job/hakuna%20matata

igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/InfoController.groovy
```diff
@@ -73,11 +76,31 @@ class InfoController {
...
+            }
+        }
+        recursiveGetJobs(jenkinsClient.jobs.list)
+
+        return jobList
     }
 
-    @RequestMapping(value = '/jobs/{master}/{job:.+}')
-    JobConfig getJobConfig(@PathVariable String master, @PathVariable String job) {
+    @RequestMapping(value = '/jobs/{master}/**')
```
AMeng: In order to allow forward slashes in the names of Jenkins jobs, we can manually parse the URL like this. If this looks acceptable, I can do something similar in the `BuildController`. I'll have to move the job names to the end of all those URLs and manually parse them in this same way.

AMeng:
To be clear, these are the URL changes I've made. Note that the job name *has* to be at the end of the URL in order to allow any arbitrary number of forward slashes in the name.

| Old URL                                                  | New URL                                                    |
|----------------------------------------------------------|------------------------------------------------------------|
| /jobs/{master}/{job}/{buildNumber}                       | /builds/status/{buildNumber}/{master}/{job}                |
| /jobs/{master}/queue/{item}                              | /builds/queue/{master}/{item}                              |
| /jobs/{master}/{job}/builds                              | /builds/all/{master}/{job}                                 |
| /jobs/{master}/{job}/{buildNumber}/properties/{fileName} | /builds/properties/{buildNumber}/{fileName}/{master}/{job} |

tomaslin:
changes look sensible. Upstream services would need to change to reflect these -

ajordens:
This looks to me like a breaking API change? 

If so, we'll need to version the API in here (or perhaps just leave old and new running side-by-side).

AMeng:
@ajordens, doesn't Gate handle talking to Igor? It seems like users shouldn't be working directly with Igor's API. I guess it makes sense to version it either way. Gate's URLs are changing as well (https://github.com/spinnaker/gate/pull/161), so we'll need to address that issue. But yes, this is a breaking API change. If you'd like to version the API, how would you prefer to handle that? Maybe just prepend the new URLs with `/v2` or something? And should we also keep the old endpoints?

tomaslin:
if we leave old and new running side by side, the UI will not be able to handle jobs with the folders plugin. My vote is to change the gate API so that it has v2 attached to it. 

Deck would need to be updated as well to support the new url patterns https://github.com/spinnaker/deck/blob/master/app/scripts/modules/core/ci/jenkins/igor.service.js

It also seems that bakery parses jobs incorrectly - https://github.com/spinnaker/orca/blob/master/orca-core/src/main/groovy/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.groovy#L120-L124 so this would have to change as well.

The last bit to sanity check is how things are constructed to link back to the build in the clusters view in deck. I suspect this might be fine but basically we need to wire up all those pieces and test it works correctly.

tomaslin:
might be useful to note that I'm just launching a new jenkins via ```docker run -p 8080:8080 -p 50000:50000 jenkins``` and then installing the folders plugin after creating a dummy user. However, since front50 runs in 8080 I change that port in my docker compose config

gregturn:
If encoding paths/filenames is an issue, is it possible to switch to making them query parameters? I think that would encode things. Not 100% sure how to run that through retrofit. I know Spring MVC encodes stuff, but I can't say the same for retrofit. If you're breaking the API anyway, this might be the least risky way to get the entire value over the wire.

AMeng:
Updated gate to prepend the new URLs with `/v2`: https://github.com/spinnaker/gate/pull/161
Updated orca to fix the Jenkins URL parse issue: https://github.com/spinnaker/orca/pull/669

tomaslin:
Deck code here - https://github.com/spinnaker/deck/pull/1766

tomaslin:
For jobs that don't have a path, the endpoint to retrieve builds is broken. ( might be just in deck )

before:

https://spinnaker/builds/spinnaker/jobs/SPINNAKER-package-igor/builds

after:

http://spinnaker/v2/builds/spinnaker/builds/jobs/SPINNAKER-package-igor

igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/InfoController.groovy
tomaslin: would probably prefer to move job name as a query parameter like some other people have suggested in the comments.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/client/JenkinsClient.groovy
```diff
@@ -40,7 +40,11 @@ interface JenkinsClient {
     @GET('/api/xml?tree=jobs[name,lastBuild[actions[failCount,skipCount,totalCount,urlName],duration,number,timestamp,result,building,url]]&exclude=/*/*/*/action[not(totalCount)]')
     ProjectsList getProjects()
 
-    @GET('/api/xml?tree=jobs[name]')
+    /*
+     * Jobs created with the Jenkins Folders plugin are nested.
+     * This query looks for jobs within folders with a depth of 10.
+     */
+    @GET('/api/xml?tree=jobs[name,jobs[name,jobs[name,jobs[name,jobs[name,jobs[name,jobs[name,jobs[name,jobs[name,jobs[name]]]]]]]]]]')
```
tomaslin: in the ideal world this kind of structure and resolution would be configurable at the jenkins masters level.

tomaslin:
Getting job configuration from deck UI does not work for folders.

Create a job at the level folder1/folder2/folderjob

Make the job parametrized and add a few parameters to the job.

In deck, create a new application -> new pipeline -> new jenkins stage

Select the master for the folder job and select this job.

Seems like the slashes get encoded when trying to get the job configuration:

http://192.168.99.100:8084/v2/builds/folders/jobs/folder1%2Fjob%2Ffolder2%2Fjob%2Ffolderjob 

The result of this is that parameters for the job are not shown

tomaslin:
Polling is broken for jobs coming from the folder plugin.

The BuildMonitor in igor gets the list of jobs via the getProjects endpoint in JenkinsClient. This information is then used to resolve if there is a new build or not and new events get sent to Echo to trigger pipelines.

https://github.com/spinnaker/igor/blob/master/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/BuildMonitor.groovy#L125

The path is here - https://github.com/spinnaker/igor/blob/78421f8cf3ba498d316338dcd173f7d3975259c3/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/client/JenkinsClient.groovy#L40-L41

Seems that for jobs with folders attached to it, only the top level folder is displayed as an available job. 

The same change to get a list of jobs needs to be made so that new build events from jobs get logged correctly.

AMeng:
I've addressed the issue with polling jobs. While testing, I also noticed that job names with spaces were causing issues because I had marked the URI as encoded to avoid encoding the forward slashes. I'll need to look over the other PRs to ensure special characters like spaces are correctly encoded.

AMeng:
Other PRs have been updated to correctly handle spaces in job names. I believe this is again ready for review.

tomaslin:
do you mind squashing your commits?

AMeng:
Squashed and rebased all PRs

ajordens:
@tomaslin I know we talked about this ... but this is a breaking change and the internal Spinnaker callers of igor (gate and orca) are going to break (even if for only an optimal 5 minutes as everything is rolled out).

@AMeng I realize the there are PRs against orca and gate, but we deploy each service independently and strive to avoid dependent rollouts.

@tomaslin What do you figure the LOE would be to re-insert the old igor endpoints? As simple as restoring the code?

AMeng:
If it is a requirement that these changes be deployable separately, then Gate will have the same issues. I've prepended the new gate URLs with `/v2`, but did not keep the old ones around. So I would need to allow the original endpoints for both Gate and Igor.

Is this a requirement?

AMeng:
@ajordens I've re-added the original endpoints for both Igor and Gate. I tried to make it as easy as possible to rip them back out whenever you think its ok to do so (both the methods and the tests are grouped on the bottom).

tomaslin:
I've tested this PR and with the other changes against a jenkins slave without the folders plugin and one with it. Job names with and without spaces work correctly in both and triggering works as well.. LGTM.

@ajordens can you run through the API changes and merge if you're satisfied?

PR #47, opened 2015-11-25
send echo events for all builds polled items

tomaslin:


tomaslin:
@spinnaker/netflix-reviewers - PTAL

PR #48, opened 2015-11-28
upgrade gradle plugin to fix idea generation (gradle 2.9)

cfieber:


ajordens:
LGTM, fixes the issue I was having with idea plugin.

PR #49, opened 2015-11-30
Introduced Hystrix around the calls to jenkins shards

ajordens:
- Each Jenkins shard gets their own hystrix pool
- No fallback behavior (currently), rather all hystrix activity will result in a 429 being raised

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/IgorConfig.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/IgorConfig.groovy
@@ -16,15 +16,20 @@
 
 package com.netflix.spinnaker.igor.config
 
+import com.netflix.hystrix.exception.HystrixRuntimeException
 import groovy.transform.CompileStatic
+import org.springframework.http.HttpStatus
+import org.springframework.web.bind.annotation.ControllerAdvice
+import org.springframework.web.bind.annotation.ExceptionHandler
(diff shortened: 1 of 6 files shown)


ajordens:
Expectation is that each jenkins shard will land in it's own Hystrix pool, and issues with a single jenkins shard will not flip a circuit breaker affecting all shards.

Currently, there is no actual fallbacks implemented. `igor` does have explicit 404 handling for missing jenkins shards that exists outside of Hystrix. 

`gate` will require an update to pass the underlying 429 through (rather than treating it as a failure).

PTAL @spinnaker/netflix-reviewers

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/IgorConfig.groovy
```diff
@@ -36,4 +41,32 @@ class IgorConfig extends WebMvcConfigurerAdapter {
...
+                failureCause = failureCause.cause ?: failureCause
+            }
+
+            return [
+                fallbackException: exception.fallbackException.toString(),
+                failureType: exception.failureType,
+                failureCause: failureCause.toString(),
+                error: "Hystrix Failure",
+                message: exception.message,
+                status: 429,
```
cfieber: `HttpStatus.TOO_MANY_REQUESTS`

cfieber:
LGTM

PR #53, opened 2015-12-02
Isolate test case with confined set of components

gregturn:


gregturn:
@cfieber Can you review this?

tomaslin:
LGTM

PR #63, opened 2016-02-05
allow timeout for jenkins http client to be modified by a property

tomaslin:
you can now set a client.timeout value in your igor.yml file and it will override the default value of 30 seconds. 

Note: the time is in milliseconds so ```client.timeout: 60000```

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/JenkinsConfig.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/JenkinsConfig.groovy
@@ -20,10 +20,12 @@ import com.netflix.spinnaker.igor.jenkins.client.JenkinsClient
 import com.netflix.spinnaker.igor.jenkins.client.JenkinsMasters
 import com.netflix.spinnaker.igor.jenkins.service.JenkinsService
 import com.squareup.okhttp.Credentials
+import com.squareup.okhttp.OkHttpClient
 import groovy.transform.CompileStatic
 import groovy.util.logging.Slf4j
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
+import org.springframework.beans.factory.annotation.Value
 import retrofit.Endpoints
 import retrofit.RequestInterceptor
 import retrofit.RestAdapter
@@ -31,6 +33,7 @@ import retrofit.client.OkClient
 import retrofit.converter.SimpleXMLConverter
 
 import javax.validation.Valid
+import java.util.concurrent.TimeUnit
 
 /**
  * Converts the list of Jenkins Configuration properties a collection of clients to access the Jenkins hosts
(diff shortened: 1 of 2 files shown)


tomaslin:
fixes https://github.com/spinnaker/spinnaker/issues/719

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/JenkinsConfig.groovy
```diff
@@ -40,11 +43,14 @@ import javax.validation.Valid
 @CompileStatic
 class JenkinsConfig {
 
+    @Value('${client.timeout:30}')
```
cfieber: is clientTimeout seconds or millis? I'd have expected millis and if so 30 is going to be not good as a default

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/JenkinsConfig.groovy
```diff
@@ -52,11 +58,14 @@ class JenkinsConfig {
         return new JenkinsService(jenkinsHostId, jenkinsClient)
     }
 
-    static JenkinsClient jenkinsClient(String address, String username, String password) {
+    static JenkinsClient jenkinsClient(String address, String username, String password, int timeout = 30) {
+        OkHttpClient client = new OkHttpClient()
+        client.setReadTimeout(timeout, TimeUnit.SECONDS)
```
cfieber: nevermind my above comment!

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/JenkinsConfig.groovy
```diff
@@ -40,11 +43,14 @@ import javax.validation.Valid
 @CompileStatic
 class JenkinsConfig {
 
+    @Value('${client.timeout:30}')
+    int clientTimeOut
```
cfieber: s/TimeOut/Timeout

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/JenkinsConfig.groovy
tomaslin: it's in seconds ( see line 63 )

cfieber: I see below where it is explicitly seconds

Maybe change the config property to client.readTimeoutSeconds

ajordens: Connection timeouts in other parts of Spinnaker are specified in
milliseconds.

Probably worth being consistent.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/JenkinsConfig.groovy
ajordens: If we're using okHttpClient in here ... feasibility on using the `OkHttpClientConfiguration` we use elsewhere, it would result in similar configuration patterns across all the services.

```
okHttpClient:
  readTimeoutMs: 59000
  connectTimoutMs: 30000
```

PR #64, opened 2016-03-03
provider/docker: Poll docker registry via clouddriver

lwander:
@duftler


igor-web/src/main/groovy/com/netflix/spinnaker/igor/health/PollingMonitorHealth.groovy
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright 2014 Netflix, Inc.
+ *
```
lwander: Keeping the Netflix copyright since it's mostly their code in this file.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/health/PollingMonitorHealth.groovy
```diff
@@ -0,0 +1,71 @@
...
+ */
+@Component
+public class PollingMonitorHealth implements HealthIndicator {
+
+    @Autowired
+    List<PollingMonitor> pollers
+
+    @Override
+    public Health health() {
+        List<Health> healths = []
```
lwander: @tomaslin, this changes the behavior of the polling agent. Is this OK? Where is this endpoint hit from?

lwander: Basically, it checks if both the Docker and Build polling agents are running, and squashes their statuses together (see line 57).

lwander:
In order to avoid duplicating code, this PR pings the clouddriver cache to see if any of the images in the requested (clouddriver) docker registry accounts have changed since the last polling cycle. If so, that's reflected in the igor cache and sent to echo.

igor-web/config/igor.yml
```diff
@@ -10,6 +10,13 @@ jenkins:
       username: 'username'
       password: 'password'
 
+dockerRegistry:
```
duftler: Please prepare a companion PR for spinnaker/spinnaker that updates the 'unified' config to reflect this new config: https://github.com/spinnaker/spinnaker/blob/master/config/igor.yml

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/DockerRegistryConfig.groovy
```diff
@@ -0,0 +1,63 @@
...
+import retrofit.Endpoints
+import retrofit.RestAdapter
+import retrofit.client.OkClient
+
+import javax.validation.Valid
+
+@Configuration
+@ConditionalOnProperty('services.clouddriver.baseUrl')
+@CompileStatic
+class DockerRegistryConfig {
```
duftler: Indentation seems off in this class.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/DockerRegistryProperties.groovy
```diff
@@ -0,0 +1,30 @@
...
+import groovy.transform.CompileStatic
+import org.hibernate.validator.constraints.NotEmpty
+import org.springframework.boot.context.properties.ConfigurationProperties
+import org.springframework.context.annotation.Configuration
+
+@Configuration
+@CompileStatic
+@ConfigurationProperties(prefix = 'dockerRegistry')
+class DockerRegistryProperties {
+    @NotEmpty
```
duftler: Please use 2 spaces for indentation everywhere, not 4.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/DockerRegistryProperties.groovy
```diff
@@ -0,0 +1,30 @@
...
+package com.netflix.spinnaker.igor.config
+
+import groovy.transform.CompileStatic
+import org.hibernate.validator.constraints.NotEmpty
+import org.springframework.boot.context.properties.ConfigurationProperties
+import org.springframework.context.annotation.Configuration
+
+@Configuration
+@CompileStatic
+@ConfigurationProperties(prefix = 'dockerRegistry')
```
duftler: Can't `prefix = ` be dropped here?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/DockerRegistryProperties.groovy
lwander: This whole component uses 4, not 2.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerRegistryCache.groovy
```diff
@@ -0,0 +1,92 @@
...
+        def split = key?.split("'")
+
+        if (!split || split[0] != prefix || split[1] != id || split.length != 6) {
+            return null
+        }
+
+        [account: split[2], registry: split[3], repository: split[4], tag: split[5]]
+    }
+
+    static public String formatImage(String registry, String repository, String tag) {
```
duftler: Can drop `public`.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerRegistryCache.groovy
```diff
@@ -0,0 +1,92 @@
...
+        registry + '/' + repository + ':' + tag // for some reason groovy string magic doesn't work here.
```
duftler: Not sure what you mean it 'doesn't work here'. If you tried to use the resulting value as a map key, and a subsequent lookup failed in that map failed, it's likely because the resulting value is not of type `String` (it will be a `GString`).
Try this: `"$registry/$repository:$tag".toString()"`

duftler: http://docs.groovy-lang.org/latest/html/documentation/index.html#_gstring_and_string_hashcodes

igor-web/src/main/groovy/com/netflix/spinnaker/igor/health/PollingMonitorHealth.groovy
```diff
@@ -0,0 +1,71 @@
...
+        pollers.forEach { poller ->
+            if (poller.isInService()) {
+                if (poller.lastPoll == null) {
+                    healths << Health.unknown().withDetail('${poller.name}.status', 'not polling yet').build()
+                } else {
+                    if (System.currentTimeMillis() - poller.lastPoll > (poller.pollInterval * 2 * 1000)) {
```
duftler: Maybe use constants here?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/DockerRegistryConfig.groovy
duftler: Ah, now I see that you were just trying to fit in with the neighbors. I'm going to delete my indentation-related comments.

duftler:
This looks reasonable to me (after addressing the minor comments). Defer to @tomaslin for comments and merging.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerRegistryCache.groovy
lwander: It's not a type error, it just returns the literal "$registry/$repository:$tag", without subsitution. Do you mean the function return type coerces this to a String before the GString substitution takes place? I'll see if .toString() makes it clearer to the Groovy compiler.

lwander: That did the trick

igor-web/src/main/groovy/com/netflix/spinnaker/igor/health/PollingMonitorHealth.groovy
tomaslin: the health endpoint is hit from /health and is used by load balancers as a signal that everything is alright.

igor-web/config/igor.yml
tomaslin: we shouldn't have to update docker registry accounts in two places, ideally the endpoint from clouddriver should be able to return all details.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/health/PollingMonitorHealth.groovy
lwander: Are the details fields consumed?

tomaslin:
there should be the ability to enable / disable a poller now that we're not doing strict jenkins polling, otherwise LGTM. 

Still need to stand it up and see that it is working properly

igor-web/config/igor.yml
lwander: Good idea. I'll add that to clouddriver then update this PR.

lwander: Actually, I'll just hit the `/credentials` endpoint.

lwander:
@tomaslin What should the health check report if no polling agents are enabled?

lwander:
I ask because right now it will show `Health.UNKNOWN` if any of them are disabled. Should we say it's up as long as one is running, and none are down?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/health/PollingMonitorHealth.groovy
tomaslin: I don't think so, mostly the status code

tomaslin: the fields are there to help humans

tomaslin:
if two pollers are enabled but only one or none running, it should report that is unhealthy.

tomaslin:
we want to respawn the service if for some reason a poller dies

lwander:
Cool, I left it as is then. It's ready to be tried out if you like, you just need to add

```
dockerRegistry:
  enabled: true
```

to igor-web/config/igor.yml file. 

I'm leaving it disabled by default for now.

tomaslin:
I'm happy with this if you have no further additions

tomaslin:
probably also want to ensure the clouddriver - igor dependency gets captured in the experimental/docker-compose configuration and the kubernetes config. Viel Glück

PR #65, opened 2016-03-05
provider/docker: Send proper event details to echo.

lwander:
I was incorrectly using the `details` field of the event data sent to Echo rather than `content`. 

I also switched formatting of `imageId` from `$registry/$repository:$tag` to the image key stored in redis, for simplicity.

@duftler

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerMonitor.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerMonitor.groovy
@@ -19,7 +19,7 @@ package com.netflix.spinnaker.igor.docker
 import com.netflix.spinnaker.igor.docker.model.DockerRegistryAccounts
 import com.netflix.spinnaker.igor.docker.service.TaggedImage
 import com.netflix.spinnaker.igor.history.EchoService
-import com.netflix.spinnaker.igor.history.model.BuildDetails
+import com.netflix.spinnaker.igor.history.model.DockerEvent
 import com.netflix.spinnaker.igor.polling.PollingMonitor
 import groovy.util.logging.Slf4j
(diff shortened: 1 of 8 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerMonitor.groovy
```diff
@@ -32,6 +32,7 @@ import rx.Scheduler
 import rx.functions.Action0
 import rx.schedulers.Schedulers
 
+import javax.print.Doc
```
duftler: s/javax.print.Doc/javax.swing.plaf.metal.OceanTheme

igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerRegistryCache.groovy
```diff
@@ -38,14 +39,7 @@ class DockerRegistryCache {
...
 
-        def res = resource.keys(key).collect {
-            def parse = parseKey(it)
-            if (parse) {
-                formatImage(parse.registry, parse.repository, parse.tag)
-            } else {
-                null
-            }
-        }.findAll { it != null }
+        List<String> res = resource.keys(key).findAll { it != null } as List
```
duftler: Could also use:

`res = resource.keys(key).findResults( { it } )`

or

`res = resource.keys(key) - null`

duftler:
LGTM after addressing comments.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerRegistryCache.groovy
tomaslin: or resource.keys(key).removeAll([null]) which might be slightly more readable

PR #72, opened 2016-03-16
Adding support for igor to stop and cancel jenkins jobs

tonyrzhang:
Hello @robfletcher @ajordens @tomaslin !

This PR is to add the ability for the jenkins stage to be cancellable. The change involves changes on both Orca and Igor.

- Orca: Jenkins stage now implements the `CancellableStage` interface and there is now a new task that hits a `stop` endpoint on Igor
- Igor: There are two different apis for canceling a jenkins job. The first is `/job/{jobName}/{buildNumber}/stop` which will stop a job that is currently running. The other one is `/queue/cancelItem` which will remove a queued job. There is also some weird jenkins api issues I ran into that I commented on inline.
- Deck: I thought about adding some sort of flag on the jenkins stage that would allow the user to opt into the cancellable aspect of the stage, but the more I thought about it the more it seemed like a bug that it didn't cancel in the first place. Let me know what you guys think about this. 

[Orca PR](https://github.com/spinnaker/orca/pull/764)

Thanks!

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/BuildController.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/BuildController.groovy
@@ -115,7 +115,7 @@ class BuildController {
         def jenkinsService = masters.map[master]
 
         // Jobs that haven't been started yet won't have a buildNumber
-        // (They're still in the queue. We use 0 to denote that case
+        // (They're still in the queue). We use 0 to denote that case
         if (buildNumber != 0) {
             jenkinsService.stopRunningBuild(jobName, buildNumber)
         }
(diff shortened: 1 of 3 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/BuildController.groovy
```diff
@@ -101,6 +101,36 @@ class BuildController {
...
+        if (buildNumber != 0) {
+            jenkinsService.stopRunningBuild(jobName, buildNumber)
+        }
+
+        // The jenkins api for removing a job from the queue (http://<Jenkins_URL>/queue/cancelItem?id=<queuedBuild>)
+        // always returns a 404. This try catch block insures that the exception is eaten instead
+        // of being handled by the handleOtherException handler and returning a 500 to orca
+        try {
+            jenkinsService.stopQueuedBuild(queuedBuild)
+        } catch (RuntimeException e) {
```
robfletcher: It would be good to trap a more specific exception type here – presumably it's a `RetrofitError` in which case you can even confirm the HTTP response code and re-throw for anything other than the 404.

tonyrzhang:
@robfletcher Yea that totally makes sense. Updated the catch block to reflect that. Thx

PR #77, opened 2016-04-15
Refactor build status endpoint in BuildController

gardleopard:
* Same method called for Jenkins and Travis integration
* Travis now returns scm info
* Minor update in README.md

The point of this is to make BuildController more generic, so it is easier to add new ci integrations. Also using igor classes to control what we expose to the other spinnaker services makes it easy to see what fields that are used from builds throughout the spinnaker platform.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
@@ -17,6 +17,7 @@
 package com.netflix.spinnaker.igor.build
 
 import com.fasterxml.jackson.databind.ObjectMapper
+import com.netflix.spinnaker.igor.build.model.GenericBuild
 import com.netflix.spinnaker.igor.jenkins.client.model.JobConfig
 import com.netflix.spinnaker.igor.jenkins.client.model.QueuedJob
 import com.netflix.spinnaker.igor.model.BuildServiceProvider
@@ -52,27 +53,17 @@ class BuildController {
     ObjectMapper objectMapper
 
     @RequestMapping(value = '/builds/status/{buildNumber}/{master}/**')
-    Map getJobStatus(@PathVariable String master, @PathVariable Integer buildNumber, HttpServletRequest request) {
+    GenericBuild getJobStatus(@PathVariable String master, @PathVariable Integer buildNumber, HttpServletRequest request) {
         def job = (String) request.getAttribute(
             HandlerMapping.PATH_WITHIN_HANDLER_MAPPING_ATTRIBUTE).split('/').drop(5).join('/')
-        if (buildMasters.filteredMap(BuildServiceProvider.JENKINS).containsKey(master)) {
-            Map result = objectMapper.convertValue(buildMasters.map[master].getBuild(job, buildNumber), Map)
+        if (buildMasters.map.containsKey(master)) {
+            GenericBuild build = buildMasters.map[master].getGenericBuild(job, buildNumber)
             try {
-                Map scm = objectMapper.convertValue(buildMasters.map[master].getGitDetails(job, buildNumber), Map)
-                if (scm?.action?.lastBuiltRevision?.branch?.name) {
-                    result.scm = scm?.action.lastBuiltRevision
-                    result.scm = result.scm.branch.collect {
-                        it.branch = it.name.split('/').last()
-                        it
-                    }
-                }
+                build.genericGitRevisions = buildMasters.map[master].getGenericGitRevisions(job, buildNumber)
             } catch (Exception e) {
                 log.error("could not get scm results for $master / $job / $buildNumber")
             }
-            result
-        } else if (buildMasters.filteredMap(BuildServiceProvider.TRAVIS).containsKey(master)) {
-            Map result = objectMapper.convertValue(buildMasters.map[master].getGenericBuild(job, buildNumber), Map)
-            result
+            return build
         } else {
             throw new MasterNotFoundException("Master '${master}' not found")
         }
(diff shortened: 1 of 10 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/model/GenericBuild.groovy
```diff
@@ -16,23 +16,37 @@
...
 
+import com.fasterxml.jackson.annotation.JsonInclude
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
 class GenericBuild {
     boolean building
     String fullDisplayName
     int number
-    int duration
+    Integer duration
```
lwander: I'm curious why you switched the type here

gardleopard: Its not something I thought strongly about. It was just to make it work and I didn't see it when I cleaned up, I will change it back.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/client/model/Build.groovy
```diff
@@ -56,6 +58,11 @@ class Build {
      */
     @ElementList(required = false, name = "action", inline = true)
     List<TestResults> testResults
+
+    GenericBuild genericBuild(String jobName) {
```
lwander: Is there a reason why you don't have Build implement the GenericBuild interface, rather than write these conversion methods? There's nothing wrong with the conversion methods, it just feels a little strange

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
```diff
@@ -55,24 +55,17 @@ class BuildController {
...
-                Map scm = objectMapper.convertValue(buildMasters.map[master].getGitDetails(job, buildNumber), Map)
-                if (scm?.action?.lastBuiltRevision?.branch?.name) {
-                    result.scm = scm?.action.lastBuiltRevision
-                    result.scm = result.scm.branch.collect {
-                        it.branch = it.name.split('/').last()
-                        it
-                    }
+                def scm = objectMapper.convertValue(buildMasters.map[master].getGenericGitRevision(job, buildNumber), Map)
+                if (scm) {
+                    result.scm = scm
```
lwander: This check is a little redundant. `if (scm)` fails, you leave `result.scm = null`, which is equivalent to just assigning `result.scm = scm` regardless of the value of `scm`.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/model/GenericBuild.groovy
gardleopard: Fixed in latest commit.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
gardleopard: I tested this with
```
result.scm = objectMapper.convertValue(buildMasters.map[master].getGenericGitRevision(job, buildNumber), Map)
```

and got 
```
 ➜ curl http://localhost:8080/builds/status/1/dummy/build-my-project
{"building":false,"fullDisplayName":"build-my-project #1","number":1,"duration":52,"timestamp":"1456914302615","result":"SUCCESS","url":"http://54.194.40.92:8080/job/build-my-project/1/","scm":null}
```

I think this is because we are using the map type on this. I can probably return a GenericBuild from this endpoint and get the code nicer.

lwander: Ah, so `scm = null` is no good in the response? No need to stress about this, we can leave it as is

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/model/GenericBuild.groovy
gardleopard: The reason I changed duration from int to Integer is that an integer does not report as 0 if it is unset. The change back to int broke the build because a test expect only building and number to be set
```
    |        |               {"building":false,"number":123(,"duration":0)}
    |        |               {"building":false,"number":123(-------------)}
    |        {"building":false,"number":123,"duration":0}
```

lwander: Then Integer is the way to go

igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/client/model/Build.groovy
gardleopard: Yes, GenericBuild is not an interface it is meant as being the build representation of igor. This means it should not be connected to the jenkins representation of a build. It just happens to be very similar because the jenkins representation of a build is whats integrated against in the rest of the services in spinnaker.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/model/ScmDetails.groovy
```diff
@@ -32,6 +32,13 @@ import org.simpleframework.xml.Root
 class ScmDetails {
     @Element(required = false)
     Action action
+
+    GenericGitRevision genericGitRevision() {
+        if (action?.lastBuiltRevision?.branch?.name) {
+            return new GenericGitRevision(action.lastBuiltRevision.branch.first().name, action.lastBuiltRevision.branch.first().name.split('/').last(), action.lastBuiltRevision.branch.first().sha1)
```
lwander: This line here: https://github.com/spinnaker/igor/pull/77/files#diff-cdf1a482044f79d54f93d79181d4b209L64 seems to modify everything in the `branch` attribute, but here you only peel off the `first()` object in `branch`. Are you sure the behavior is identical before and after your changes?

PR #78, opened 2016-04-15
Travis: synchronize repositories and validate the build cache

gardleopard:
* Spawn separate worker thread for consistency work
* Make a travis call to sync travis against git
* Validate and delete jobs that are in the build cache and not in travis

The workflow we are implementing for using travis with spinnaker at schibsted is as follows.
1. We have a spinnaker system user in github
2. To get your builds into spinnaker, you grant the spinnaker system user access to the repository

Now igor syncs travis and github for us and makes new builds appear at in average 2 minutes and 30 seconds. The sync interval can be changed by the spinnaker.build.travis.repositorySyncInterval property.

The reason this is done in a separate thread is that the build cache validation can be quite expensive in large installations.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.groovy
@@ -153,7 +153,7 @@ class TravisBuildMonitor implements PollingMonitor{
         startTime = System.currentTimeMillis()
         Observable.from(cachedRepoSlugs).subscribe(
             { String repoSlug ->
-                if(!buildMasters.map[master].getRepo(repoSlug)) {
+                if(!buildMasters.map[master].hasRepo(repoSlug)) {
                     log.info "repositorySync: Removing ${master}:${repoSlug} from buildCache because it is not on ${master} anymore."
                         buildCache.remove(master, repoSlug)
                 }
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.groovy
@@ -56,6 +56,8 @@ class TravisBuildMonitor implements PollingMonitor{
 
     Scheduler.Worker worker = Schedulers.io().createWorker()
 
+    Scheduler.Worker repositorySyncWorker = Schedulers.io().createWorker()
+
     @Autowired
     BuildCache buildCache
 
@@ -70,11 +72,17 @@ class TravisBuildMonitor implements PollingMonitor{
 
     Long lastPoll
 
+    Long lastRepositorySync
+
     static final String BUILD_IN_PROGRESS = 'started'
 
     @SuppressWarnings('GStringExpressionWithinString')
     @Value('${spinnaker.build.pollInterval:60}')
     int pollInterval
+
+    @SuppressWarnings('GStringExpressionWithinString')
+    @Value('${spinnaker.build.travis.repositorySyncInterval:300}')
+    int repositorySyncInterval
 
     @Override
     void onApplicationEvent(ContextRefreshedEvent event) {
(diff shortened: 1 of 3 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.groovy
```diff
@@ -92,6 +100,19 @@ class TravisBuildMonitor implements PollingMonitor{
                 }
             } as Action0, 0, pollInterval, TimeUnit.SECONDS
         )
+
+        worker.schedulePeriodically(
+            {
+                if (isInService()) {
```
lwander: Can you be sure that these workers won't step on each other if the sync interval is too low? Maybe protect the `repositorySync` with a semaphore, and abort the sync if you can't get a hold of the semaphore.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/service/TravisService.groovy
```diff
@@ -69,6 +69,17 @@ class TravisService implements BuildService{
         accounts
     }
 
+    void setAccounts() {
+        this.accounts = getAccounts()
+    }
+
+    void syncRepos() {
+        Response response = travisClient.usersSync(getAccessToken())
+        if (response.status != 200) {
```
lwander: 200 is typically not the only success code. Usually 200 <= status < 300 are considered OK.

lwander: I'm sure retrofit has some method on the `Response` object that will check for success. It might even throw an exception rather than return if there was an error.

lwander:
@tomaslin can you take a look at this as well?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/service/TravisService.groovy
gardleopard: @lwander I checked retrofit, I could not find a function like that. I changed the test to be >= 400. That should cover the errors.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.groovy
gardleopard: I've had it running for 9 hours with this code now without failure. I can add a semaphore, but then it might just as well be the changedBuilds method that calls repositorySync every 10th time it runs or something. What do you think?
1. Leave it as is.
2. Introduce semaphore.
3. Let changedBuilds call repositorySync for every X time it runs.

tomaslin:
it's not clear to me why this needs to be a functionality that igor provides

igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.groovy
tomaslin: another way to ensure atomicity is to use lua scripts and let redis deal with it

igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/service/TravisService.groovy
```diff
@@ -70,6 +71,17 @@ class TravisService implements BuildService {
         return accounts
     }
 
+    void setAccounts() {
+        this.accounts = getAccounts()
+    }
+
+    void syncRepos() {
+        Response response = travisClient.usersSync(getAccessToken())
```
ajordens: This won't actually return a 400 ... it'll throw a RetrofitException.

gardleopard:
Thanks for all the feedback guys. I have spoken with @tomaslin on slack in addition to the comments here. I will close this PR while I work on a slightly different solution. I will make a toggle flag for the syncRepos call and address the remaining comments. When I have a working solution I will open this PR again.

PR #79, opened 2016-04-15
Property endpoint for travis is not working, added fix for it

gardleopard:
* Fix travis property endpoint
* Add test for travis property endpoint
* add explicit return to methods
* add space around if statements
* remove some unused code


ajordens:
LGTM

PR #80, opened 2016-04-16
Add support for triggering travis builds with igor

gardleopard:
* https://travis-ci.org confirmed working
* https://travis-ci.com (travis pro) @aioue will test it on travis-pro.
* Does not work on travis enterprise (travis enterprise is way behind the other versions)

The endpoint we are using is in beta. https://docs.travis-ci.com/user/triggering-builds

The builds in the screenshot are triggered from travis using 
`curl -s -X PUT http://localhost:8080/masters/travis-ci/jobs/repo/slug`

screen shot 2016-04-16 at 22 10 27

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisCache.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisCache.groovy
@@ -1,0 +1,86 @@
+/*
+ * Copyright 2016 Schibsted ASA.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.igor.travis
+
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.stereotype.Service
+import redis.clients.jedis.Jedis
+import redis.clients.jedis.JedisPool
+
+
+/*
+    This creates an internal queue for igor triggered travis jobs.
+    The travis api does not return an id we can track in the queue.
+ */
+@Service
+@ConditionalOnProperty("travis.enabled")
+class TravisCache {
+    @Autowired
+    JedisPool jedisPool
+
+    @SuppressWarnings('GStringExpressionWithinString')
+    @Value('${spinnaker.jedis.prefix:igor}')
+    String prefix
+
+    String id = 'travis:builds:queue'
+
+    Map getQueuedJob(String master, int queueNumber) {
+        Jedis resource = jedisPool.getResource()
+
+        resource.withCloseable {
+            Map result = resource.hgetAll(makeKey(master, queueNumber))
+            if (!result) {
+                return [:]
+            }
+            Map convertedResult = [
+                buildNumber: Integer.parseInt(result.buildNumber),
+                jobName: result.jobName
+            ]
+            return convertedResult
+        }
+    }
+
+    int setQueuedJob(String master, String jobName, int buildNumber) {
+        Jedis resource = jedisPool.getResource()
+        int queueId = (int) (long) new Date().getTime()
+        resource.withCloseable {
+            while (resource.exists(makeKey(master, queueId))) {
+                queueId++;
+            }
+            resource.hset(makeKey(master, queueId), 'buildNumber', buildNumber as String)
+            resource.hset(makeKey(master, queueId), 'jobName', jobName)
+            return queueId
+        }
+    }
+
+    void remove(String master, int queueId) {
+        Jedis resource = jedisPool.getResource()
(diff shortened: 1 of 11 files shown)


gardleopard:
This successfully triggers jobs in travis, however it does not enable spinnaker to track the build after its started so the stage turn red even if the build succeeds. I will do more work on this in order to get it working like jenkins does.

gardleopard:
Got it working now. There are more stuff to do before it is good for merge, but it can be tested now.

Example pipeline triggering travis job and collecting the result of it bellow:
screen shot 2016-04-17 at 22 08 59
screen shot 2016-04-17 at 22 09 05

gardleopard:
This is ready for review. Please note that it adds a travisCache to keep track of the triggered build in travis. Travis trigger build endpoint does not return an id that we can use to track the build we triggered, igor needs to guess what build that it triggered.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisCache.groovy
```diff
@@ -0,0 +1,83 @@
...
+@ConditionalOnProperty("travis.enabled")
+class TravisCache {
+    @Autowired
+    JedisPool jedisPool
+
+    @SuppressWarnings('GStringExpressionWithinString')
+    @Value('${spinnaker.jedis.prefix:igor}')
+    String prefix
+
+    String id = 'builds:queue'
```
lwander: Maybe change to `travis:builds:queue` to be sure we don't accidentally conflict with jenkins.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisCache.groovy
```diff
@@ -0,0 +1,83 @@
...
+
+    Map getQueuedJob(String master, int queueNumber) {
+        Jedis resource = jedisPool.resource
+        if (!resource.exists(makeKey(master, queueNumber))) {
```
lwander: Can any of these jedis calls throw an exception? I might wrap these in a `try` and put the `returnResource` in a `finally` to be sure we don't leak resources.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/service/TravisResultConverter.groovy
```diff
@@ -25,6 +25,9 @@ import groovy.util.logging.Slf4j
 class TravisResultConverter {
     static Result getResultFromTravisState(String state) {
         switch (state) {
+            case "created":
+                return Result.NOT_BUILT
+                break
             case "started":
                 return Result.BUILDING
                 break
```
lwander: I still think this line https://github.com/spinnaker/igor/pull/80/files#diff-6e2e5f43c3a3dc4ecf7a723feb49ee26R47 should throw an exception, since it's a failure state for igor.

lwander: Looks like that link doesn't work, revised line 47 on this file is what I meant to point it at.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/service/TravisService.groovy
```diff
@@ -86,7 +95,7 @@ class TravisService implements BuildService {
 
     Build getBuild(String repoSlug, int buildNumber) {
         Builds builds = getBuilds(repoSlug, buildNumber)
-        return builds.builds.first()
+        return builds.builds.size() > 0 ? builds.builds.first() : null
```
lwander: Can `builds.builds` ever be `null`?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisCache.groovy
gardleopard: Done in latest commit.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisCache.groovy
gardleopard: Thanks @lwander, I solved this a little bit different in https://github.com/spinnaker/igor/pull/80/commits/1da9404c3b72b297bd8328b6139e38d2b5225728 . I surrounded the calls with Jedis.withCloseable. That should save if redis fails.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/service/TravisResultConverter.groovy
gardleopard: I've changed this so it returns IllegalArgumentException if the state is not handled by the different cases.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/service/TravisService.groovy
gardleopard: @lwander I did some tests against the travis api, changing build number to one that was to high and job name to a job that does not exist. Travis returns `{"builds":[],"commits":[]}` for both. So I think the answer is that it can't be null.

I added a test with `{"builds":[],"commits":[]}` as return value, just to have it in the tests.

gardleopard:
@lwander I have gone through all your comments now, please have a look at the changes and let me know if there is anything more I should fix here.

lwander:
LGTM. If you've verified this against Jenkins this is ready to merge after you squash.

gardleopard:
@lwander thanks for asking for the jenkins test. I revealed a bug in the jenkins integration that I had introduced. It is fixed in this PR https://github.com/spinnaker/igor/pull/81 (merged upstream). I have tested subscribing to builds from jenkins and triggering builds in jenkins, it works from this PR as well now.

PR #83, opened 2016-04-27
Sync repositories in travis for the spinnaker github user

gardleopard:
The workflow we are implementing for using travis with spinnaker at schibsted is as follows.
1. We have a spinnaker system user in github
2. To get your builds into spinnaker, you grant the spinnaker system user access to the repository

Now igor syncs travis and github for us and makes new builds appear.

### DIFF
--- a/igor-web/config/igor.yml
+++ b/igor-web/config/igor.yml
@@ -5,11 +5,14 @@ netflix:
 spinnaker:
   build:
     pollInterval: 30
-    
+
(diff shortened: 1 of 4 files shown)


gardleopard:
@tomaslin here is a solution of the sync for travis with a feature toggle in igor.yml. This feature toggle is wide, you turn it on for all travis masters at once.

tomaslin:
LGTM

PR #84, opened 2016-04-28
Refactor travis build monitor

gardleopard:
Only functional change is to not log `sending build events for builds between X and Y` when there is only one new build

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.groovy
@@ -74,6 +74,8 @@ class TravisBuildMonitor implements PollingMonitor{
 
     Long lastPoll
 
+    static final int NEW_BUILD_EVENT_THRESHOLD = 1
+
     static final String BUILD_IN_PROGRESS = 'started'
 
     @SuppressWarnings('GStringExpressionWithinString')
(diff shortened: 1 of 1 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.groovy
```diff
@@ -230,4 +195,50 @@ class TravisBuildMonitor implements PollingMonitor{
...
+
+            }
+        }
+    }
+
+    private void pushEventsForMissingBuilds(Repo repo, Map cachedBuild, String master, TravisService travisService) {
+        int currentBuild = repo.lastBuildNumber
+        int lastBuild = Integer.valueOf(cachedBuild.lastBuildLabel)
+
+        if (lastBuild+1 < currentBuild) {
```
duftler: This routine might read easier if you were to capture the 'build event threshold" as a constant and then create a local variable for the `lastBuild + newBuildEventThreshold` (that might not be a great name, but you get the idea...). It would let you avoid doing the arithmetic twice and would identify the `1` as some meaningful (configurable) value.

duftler:
LGTM after addressing the one comment.

duftler:
The overall Travis integration is really looking great.

Maybe almost time for an announcement on the spinnaker site and an accompanying blog post?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.groovy
gardleopard: Thanks, fixed and squashed.

PR #86, opened 2016-04-28
Be less sensitive to partially-specified jenkins and travis config wh…

duftler:
…en the respective integrations are disabled.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/InfoController.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/InfoController.groovy
@@ -45,27 +45,27 @@ class InfoController {
     @Autowired
     BuildCache buildCache
 
-    @Autowired
+    @Autowired(required = false)
     JenkinsProperties jenkinsProperties
 
     @Autowired
     BuildMasters buildMasters
 
-    @Autowired
+    @Autowired(required = false)
     TravisProperties travisProperties
 
     @RequestMapping(value = '/masters', method = RequestMethod.GET)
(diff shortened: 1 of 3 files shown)


duftler:
@ewiseblatt please review.
@tomaslin fyi
@gardleopard fyi

duftler:
Had a build failure last night where igor would not start and yielded spring errors:
```
Field error in object 'travis' on field 'masters[0].githubToken': rejected value []; codes [NotEmpty.travis.masters[0].githubToken,NotEmpty.travis.masters.githubToken,NotEmpty.masters[0].githubToken,NotEmpty.masters.githubToken,NotEmpty.githubToken,NotEmpty.java.lang.String,NotEmpty]; arguments [org.springframework.context.support.DefaultMessageSourceResolvable: codes [travis.masters[0].githubToken,masters[0].githubToken]; arguments []; default message [masters[0].githubToken]]; default message [may not be empty]
```

The unified config had travis disabled, but still specified a travis master (which should be just fine): https://github.com/spinnaker/spinnaker/blob/master/config/default-spinnaker-local.yml#L162-L167.

But, all the attributes of `TravisProperties` are required (and the sample config does not specify all the properties): https://github.com/spinnaker/igor/blob/master/igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/TravisProperties.groovy#L34-L43

We shouldn't require the user to also comment-out config; just disabling the integrations should be sufficient.

This PR makes it so `JenkinsProperties` and `TravisProperties` are not populated at all when the integrations are disabled.

gardleopard:
LGTM

ewiseblatt:
LGTM

PR #88, opened 2016-05-02
Include travis master in log if repository synchronization fails

gardleopard:
The groupKey does not need to prefix with `travis-`. It is already done here:
https://github.com/spinnaker/igor/blob/master/igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/TravisConfig.groovy#L63

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/service/TravisService.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/service/TravisService.groovy
@@ -53,7 +53,7 @@ class TravisService implements BuildService {
     private Accounts accounts
 
     TravisService(String travisHostId, String baseUrl, String githubToken, TravisClient travisClient, TravisCache travisCache) {
-        this.groupKey     = "travis-${travisHostId}"
+        this.groupKey     = "${travisHostId}"
         this.githubToken  = githubToken
         this.travisClient = travisClient
         this.baseUrl      = baseUrl
@@ -270,7 +270,7 @@ class TravisService implements BuildService {
         try {
             travisClient.usersSync(getAccessToken())
         } catch (RetrofitError e) {
-            log.error "synchronizing travis repositories failed with error: ${e.message}"
+            log.error "synchronizing travis repositories for ${groupKey} failed with error: ${e.message}"
         }
     }
 


igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/service/TravisService.groovy
```diff
@@ -53,7 +53,7 @@ class TravisService implements BuildService {
     private Accounts accounts
 
     TravisService(String travisHostId, String baseUrl, String githubToken, TravisClient travisClient, TravisCache travisCache) {
-        this.groupKey     = "travis-${travisHostId}"
+        this.groupKey     = "${travisHostId}"
```
lwander: Why remove the `travis-` prefix?

lwander: ah nvm you already answered that.

lwander: I would recommend keep the prefix here, and then removing it [here](https://github.com/spinnaker/igor/blob/master/igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/TravisConfig.groovy#L63), so the key names are generated consistently. Are the keys every used in caching? If so, then it's especially important so they don't clash with jenkins keys.

gardleopard: If you look [here](https://github.com/spinnaker/igor/blob/master/igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.groovy#L134), [here](https://github.com/spinnaker/igor/blob/master/igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.groovy#L136) and [here](https://github.com/spinnaker/igor/blob/master/igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.groovy#L96) you can see that the travis-${name} thing is used from the same source a lot.

I set it in the config for preventing name clash, because I introduced [BuildCache](https://github.com/gardleopard/igor/blob/groupkey_in_travis_service_fix/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildCache.groovy) which is a generic cache every new ci integration in igor should/could use.

So if I remove it from TravisConfig I must repeat the pattern in TravisService and in TravisBuildMonitor. So my idea was to have this pattern in TravisConfig so that all classes using these setting would get the `travis-` prefixed name.

If you still prefer to set the prefix in TravisBuildMonitor and TravisService, I will do it:)

lwander: Oh, I understand. Makes sense!

PR #89, opened 2016-05-04
Travis - make virtual branches for pull requests

gardleopard:
We discovered that pull request builds are registered as the branch you do the pull request against with `"pull_request":true` in the [response](https://docs.travis-ci.com/api#builds) from travis. This pull request makes igor identify these pull requests and creates virtual branches for the target branches. A pull request to master will be registered to spinnaker as the branch `pull_request_master`.

The idea behind this is that you can create pipelines in spinnaker that listens to pull requests and run more test on them than you do in travis. You can bake and deploy them to check them out.

The current code in igor will let pull requests trigger pipelines that listens on the target branch for the pull request.


gardleopard:
@spinnaker/reviewers PTAL

lwander:
LGTM, let's merge this tomorrow unless anyone has objections before then.

PR #90, opened 2016-05-04
Make redis.timeout override for igor fixes spinnaker/spinnaker#877

gardleopard:
This lets users override the redis timeout in igor.


ajordens:
LGTM

PR #95, opened 2016-05-17
Ignore travis builds that are older than the travis cache TTL

gardleopard:
If the travis api returns old builds when the travisBuildMonitor fetches builds, we need to filter them out so that we don't get another event pushed to echo after the TTL has ran out.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.groovy
@@ -74,6 +74,8 @@ class TravisBuildMonitor implements PollingMonitor{
 
     static final int NEW_BUILD_EVENT_THRESHOLD = 1
 
+    static final long BUILD_STARTED_AT_THRESHOLD = TimeUnit.SECONDS.toMillis(30)
+
     @SuppressWarnings('GStringExpressionWithinString')
     @Value('${spinnaker.build.pollInterval:60}')
     int pollInterval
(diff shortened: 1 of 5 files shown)


gardleopard:
@spinnaker/reviewers PTAL

igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.groovy
```diff
@@ -74,6 +74,8 @@ class TravisBuildMonitor implements PollingMonitor{
 
     static final int NEW_BUILD_EVENT_THRESHOLD = 1
 
+    static final long BUILD_STARTED_AT_THRESHOLD = 30000L
```
duftler: I think this might read better as `TimeUnit.SECONDS.toMillis(30)`, especially where the constant is added to the ms value of `cachedJobTTLDays`.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.groovy
```diff
@@ -254,4 +256,11 @@ class TravisBuildMonitor implements PollingMonitor{
     private int buildCacheJobTTLSeconds() {
         return TimeUnit.DAYS.toSeconds(cachedJobTTLDays)
     }
+
+    private List<Repo> filterOutOldBuilds(List<Repo> repos){
+        Long threshold = new Date().getTime() - TimeUnit.DAYS.toMillis(cachedJobTTLDays) + BUILD_STARTED_AT_THRESHOLD
```
duftler: This could benefit from a comment explaining why the threshold needs the additional offset beyond the `cachedJobTTLDays`.

duftler:
LGTM. Needs a rebase.

gardleopard:
@duftler Wow that was fast!

PR #99, opened 2016-05-25
docker: Log changes to digests

lwander:
This will simplify tracking down the spurious trigger cause, and should be logged anyhow for traceabilty into what was deployed.

@duftler


duftler:
LGTM.

PR #109, opened 2016-06-08
Travis support overriding environment variables when triggering builds

gardleopard:
This adds parameterDefinitionList to the getJobConfig endpoint so that the `Job Parameters` field in deck for setting up triggers gets populated based on the travis job configuration.
screen shot 2016-06-08 at 08 02 21

Also the triggering for travis builds now takes a map of requestParams and overrides env config in the travis build when triggering.

screen shot 2016-06-08 at 08 03 24

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/service/JenkinsService.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/service/JenkinsService.groovy
@@ -18,6 +18,7 @@
 package com.netflix.spinnaker.igor.jenkins.service
 
 import com.netflix.spinnaker.hystrix.SimpleHystrixCommand
+import com.netflix.spinnaker.igor.build.BuildController
 import com.netflix.spinnaker.igor.build.model.GenericBuild
 import com.netflix.spinnaker.igor.build.model.GenericGitRevision
 import com.netflix.spinnaker.igor.jenkins.client.JenkinsClient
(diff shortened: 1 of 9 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/client/model/Config.groovy
```diff
@@ -0,0 +1,43 @@
...
+
+    Config(Map<String, String> environmentMap) {
+        env = environmentMap.collect { key, value ->
+            "${key}=${value}".toString()
+        }
+    }
+
+    public List<GenericParameterDefinition> getParameterDefinitionList() {
+        env? env.collect {
+            def parts = it.tokenize('=')
```
lwander: https://tools.ietf.org/html/rfc3986#section-3.4

query parameters don't necessarily have `=` in them, the array access below can fail.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/service/JenkinsService.groovy
```diff
@@ -103,6 +103,12 @@ class JenkinsService implements BuildService{
         return getBuild(jobName, buildNumber).genericBuild(jobName)
     }
 
+    @Override
+    int triggerBuildWithParameters(String job, Map<String, String> queryParameters) {
+        Response response = buildWithParameters(job, queryParameters)
+        return response.body.toString().toInteger()
```
lwander: Is body always non-null? Also, will it always be an Integer?

lwander:
LGTM after comments.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/client/model/Config.groovy
gardleopard: This is not parsing query parameters, it does not take arguments. It splits the local env list that is set with equal sign either from the travis build definition or the constructor. It will break if the value contains a `=`, I will add a test for that.

lwander: Ah, I see, sounds good.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/service/JenkinsService.groovy
gardleopard: This was wrong, I have copied the code from the BuildController to extract the queued build. This method is not called as it is now, but I want to refactor so that the trigger build part in BuildController does not care whether it is jenkins or travis its triggering.
I tested some local triggering to see what the responses from igor is when triggering:
```
me@host $ curl http://172.31.0.134:8088/masters/spinnaker-jenkins/jobs/integration-tests -X PUT
168
me@host $ curl http://172.31.0.134:8088/masters/spinnaker-jenkins/jobs/whatsup?INTEGRATION_TEST_TARGET=4 -X PUT
169
```

gardleopard:
Hi @lwander, thanks for reviewing, I have addressed your comments now:)

lwander:
Looking good, I'll merge after you squash.

gardleopard:
Thanks, squashed and ready!

PR #112, opened 2016-06-16
fix poller names in health check

tomaslin:
in groovy, ${ value } within single quotes don't get evaluated, this PR renders the names of the pollers correctly in the health check


tomaslin:
@lwander FYI

lwander:
Retroactive LGTM

PR #127, opened 2016-08-10
Travis - make igor able to override multiple properties in build

gardleopard:
The travis api acts strange when we trigger builds. We need to inject one property in `env` to be able to override `global_env`. We have to override `global_env` to override multiple properties in one job, if we do that in `env` it will trigger a job in travis for every property as a build matrix. The data inside `config.global_env` can both include strings and secrets `"MYPROPERTY=foo"` and secrets in a map like `{"secure":"xyz"}`. The code in this PR gathers the secrets from the last build and injects them into the payload when we trigger builds with igor against travis.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/client/model/Config.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/client/model/Config.groovy
@@ -17,6 +17,7 @@
 package com.netflix.spinnaker.igor.travis.client.model
 
 import com.fasterxml.jackson.annotation.JsonInclude
+import com.google.gson.annotations.SerializedName
 import com.netflix.spinnaker.igor.build.model.GenericParameterDefinition
 import groovy.transform.CompileStatic
 import org.simpleframework.xml.Default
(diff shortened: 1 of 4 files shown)


gardleopard:
@spinnaker/reviewers PTAL

gardleopard:
@spinnaker/reviewers still waiting on this one

igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/client/model/Config.groovy
```diff
@@ -26,18 +27,26 @@ import org.simpleframework.xml.Default
...
-            def parts = it.tokenize('=')
+        globalEnv? globalEnv.findAll{it instanceof String}.collect {
+            String tmpGlobalEnv = (String) it
+            def parts = tmpGlobalEnv.tokenize('=')
             new GenericParameterDefinition(parts[0], parts.drop(1).join('='))
         } : []
     }
+
+    public List<Object> getSecrets() {
+        globalEnv? globalEnv.findAll{!(it instanceof String)}.collect{it} : []
```
lwander: Why are non-strings secrets?

igor-web/src/test/groovy/com/netflix/spinnaker/igor/travis/client/model/ConfigSpec.groovy
```diff
@@ -73,4 +74,21 @@ class ConfigSpec extends Specification {
...
+        config.globalEnv = globalEnv
+
+        expect:
+        config.getSecrets() == secrets
+
+        where:
+        globalEnv                          || secrets
+        [ 'FOO=bar' ]                      || []
+        [ 'FOO=bar', ["secure":"xyz"] ]    || [ [secure:"xyz"] ]
+        [ [secure:"xyz"], [secure:"abc"] ] || [ [secure:"xyz"], [secure:"abc"] ]
```
lwander: Ah I guess secrets are maps? Why is that?

lwander:
LGTM after questions

igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/client/model/Config.groovy
gardleopard: Env variables are passed as Strings, so whats not string in the list is secrets.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/travis/client/model/ConfigSpec.groovy
gardleopard: I have no idea why secrets are represented this way and I don't know why we have to send them when triggering. If we do not collect the secrets and send them, the triggered build will not have these secrets. I have reached out to travis on this and I hope they will fix it so that we do not need to send the secrets and can remove that part of the code.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/client/model/Config.groovy
```diff
@@ -26,18 +27,26 @@ import org.simpleframework.xml.Default
 @JsonInclude(JsonInclude.Include.NON_NULL)
 class Config {
 
-    List<String> env
+    @SerializedName("global_env")
+    List<Object> globalEnv
+
+    List<String> env = ["SPINNAKER_TRIGGERED=true"]
```
gardleopard: This part is also quite strange. We need to override a random env variable in order to override the global_env. Travis is looking into this as well.

PR #129, opened 2016-09-23
Handle "." in request parameter

jeyrschabu:
- Updated to use masters/{name:.+} to handle names like jenkins.net


jeyrschabu:
@ajordens PTAL . This takes care of https://github.com/spinnaker/spinnaker/issues/1165

ajordens:
I don't think it's actually necessary in all these spots, only the ones where the `{master}` would be the last element in the path?

jeyrschabu:
You are right. Did a quick test, only necessary for the last element (or potential last element) . 
This is now updated

ajordens:
LGTM

PR #132, opened 2016-10-14
Travis include artifacts in trigger event.

gardleopard:
We want to implement support for multi-repo incremental builds in spinnaker.
Given a project that creates two or more services and that is building incrementally,
we want to be able to filter in the trigger stage on which artifacts that have been uploaded.

This will let us have many pipelines listening to the same build, but only trigger them
when the artifact we will use in the bake is present.


gardleopard:
@spinnaker/reviewers PTAL

robzienert:
@gardleopard Over a year old - what's the fate here?

gardleopard:
@robzienert I will close it for now, I don't have the capacity to make the required changes in the other parts to leverage this change atm.

PR #134, opened 2016-10-26
Converts Value to ConfigurationProperties.

cfieber:
This is dependant on spinnaker/kork#62 - do not merge

Also includes:
- elimination of some Autowired fields into Configuration classes
- updating some existing ConfigurationProperties to be imported by a Configuration instead of being their own Configuration
- cleanup of some redis calls to ensure connections get closed

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,7 +23,7 @@ buildscript {
         maven { url "http://spinnaker.bintray.com/gradle" }
     }
     dependencies {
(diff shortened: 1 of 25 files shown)


ewiseblatt:
Do you want me to try this to see if it fixes the problem I was seeing in your kork PR?

cfieber:
sure (though even before this change I wasn't able to repro the problem you mentioned)

PR #135, opened 2016-10-28
Fix GetCommitList from Github. Invert from and to

gbougeard:
Should replace https://github.com/spinnaker/orca/pull/853

PTAL @dzapata @tomaslin


gbougeard:
@cfieber @ajordens PTAL

PR #138, opened 2016-11-25
Decorate artifacts in igor.

gardleopard:
This PR moves the knowledge of rpm and deb packages out of orca and into igor.
Artifact decoration can be feature toggled on and off at will.

The code here is not done yet, but the skeleton and most of the logic is present.

Currently this code only decorates rpm and deb packages. The design should make
it easy to add support for more package types/formats. Also I plan to add a parser
that can be configured through igor.yml. This would make it possible to add support
for new types of artifacts without touching the codebase in igor.

Please provide feedback on this both on the implementation and if this fits better
in orca.

There has been some concerns about property file handling. I think this PR keeps all
current functionality in igor intact.

spinnaker/spinnaker#1348

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/artifact/decorator/ConfigurableFileDecorator.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/artifact/decorator/ConfigurableFileDecorator.groovy
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2017 Schibsted ASA.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.igor.build.artifact.decorator
+
+import com.netflix.spinnaker.igor.build.model.GenericArtifact
+
+class ConfigurableFileDecorator implements ArtifactDetailsDecorator {
+
+    final String decoratorRegex
+    final String identifierRegex
+    final String type
+
+    ConfigurableFileDecorator(String type, String decoratorRegex, String identifierRegex) {
+        this.decoratorRegex = decoratorRegex
+        this.identifierRegex = identifierRegex
+        this.type = type
+    }
+
+    @Override
+    GenericArtifact decorate(GenericArtifact genericArtifact) {
+        def m
+
+        if ((m = genericArtifact.fileName =~ decoratorRegex)) {
+            genericArtifact.name = m.group(1)
+            genericArtifact.version = m.group(2)
+            genericArtifact.type = m.groupCount() > 2 ? m.group(3) : type
+            genericArtifact.reference = genericArtifact.fileName
+        }
+
+        return genericArtifact
+    }
+
+    @Override
+    boolean handles(GenericArtifact genericArtifact) {
+        return genericArtifact.fileName =~ identifierRegex
+    }
+
+    @Override
+    String decoratorName() {
+        return type
+    }
+}
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/artifact/decorator/RpmDetailsDecorator.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/artifact/decorator/RpmDetailsDecorator.groovy
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2017 Schibsted ASA.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.igor.build.artifact.decorator
+
+import com.netflix.spinnaker.igor.build.model.GenericArtifact
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.stereotype.Component
+
+@Component
+@ConditionalOnProperty('artifact.decorator.enabled')
+class RpmDetailsDecorator implements ArtifactDetailsDecorator {
+
+    String packageType = 'rpm'
+    String versionDelimiter = '-'
+
+    @Override
+    GenericArtifact decorate(GenericArtifact genericArtifact) {
+        genericArtifact.name = extractName(genericArtifact.fileName)
+        genericArtifact.version = extractVersion(genericArtifact.fileName)
+        genericArtifact.type = packageType
+        genericArtifact.reference = genericArtifact.fileName
+        return genericArtifact
+    }
+
+    @Override
+    boolean handles(GenericArtifact genericArtifact) {
+        if (!genericArtifact.fileName) {
+            return false
+        }
+        return genericArtifact.fileName.tokenize('.').last() == "rpm"
+    }
+
+    @Override
+    String decoratorName() {
+        return packageType
+    }
+
+    String extractVersion(String file) {
+        List<String> parts = file.tokenize(versionDelimiter)
+        parts.pop()
+        return parts.pop()
+    }
+
+    String extractName(String file) {
+        List<String> parts = file.tokenize(versionDelimiter)
+        parts.pop()
+        parts.pop()
+        return parts.join(versionDelimiter)
+    }
+}
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/model/GenericArtifact.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/model/GenericArtifact.groovy
@@ -16,14 +16,24 @@
 
 package com.netflix.spinnaker.igor.build.model
 
+import com.fasterxml.jackson.annotation.JsonInclude
+import groovy.transform.ToString
+
+@ToString(includeNames = true)
+@JsonInclude(JsonInclude.Include.NON_NULL)
 class GenericArtifact {
     String fileName
     String displayPath
     String relativePath
+    String reference
+    String name
+    String type
+    String version
 
     GenericArtifact(String fileName, String displayPath, String relativePath) {
         this.fileName = fileName
         this.displayPath = displayPath
         this.relativePath = relativePath
     }
+
 }
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/service/ArtifactDecorator.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/service/ArtifactDecorator.groovy
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2017 Schibsted ASA.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.igor.service
+
+import com.netflix.spinnaker.igor.build.artifact.decorator.ArtifactDetailsDecorator
+import com.netflix.spinnaker.igor.build.artifact.decorator.ConfigurableFileDecorator
+import com.netflix.spinnaker.igor.build.model.GenericArtifact
+import com.netflix.spinnaker.igor.build.model.GenericBuild
+import com.netflix.spinnaker.igor.config.ArtifactDecorationProperties
+import groovy.util.logging.Slf4j
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.stereotype.Component
+
+import javax.validation.Valid
+
+@Slf4j
+@Component
+@ConditionalOnProperty('artifact.decorator.enabled')
+class ArtifactDecorator {
+
+    List<ArtifactDetailsDecorator> artifactDetailsDecorators;
+
+    @Autowired(required = false)
+    ArtifactDecorationProperties artifactDecorationProperties
+
+    @Autowired
+    ArtifactDecorator(List<ArtifactDetailsDecorator> artifactDetailsDecorators, @Valid ArtifactDecorationProperties artifactDecorationProperties) {
+        this.artifactDetailsDecorators = artifactDetailsDecorators
+        List<ArtifactDetailsDecorator> configuredArtifactDetailsDecorators = artifactDecorationProperties?.fileDecorators?.collect { ArtifactDecorationProperties.FileDecorator fileDecorator ->
+            log.info "Configuring custom artifact decorator of type : ${fileDecorator.type}"
+            (ArtifactDetailsDecorator) new ConfigurableFileDecorator(fileDecorator.type, fileDecorator.decoratorRegex, fileDecorator.identifierRegex)
+        }
+        if (configuredArtifactDetailsDecorators) {
+            this.artifactDetailsDecorators.addAll(0, configuredArtifactDetailsDecorators)
+        }
+    }
+
+    void decorate(GenericArtifact genericArtifact) {
+        ArtifactDetailsDecorator artifactDetailsDecorator = artifactDetailsDecorators.find { it.handles(genericArtifact) }
+
+        if (artifactDetailsDecorator) {
+            artifactDetailsDecorator.decorate(genericArtifact)
+            log.debug "Decorated artifact with decorator [${artifactDetailsDecorator.decoratorName()}]: ${genericArtifact.toString()}"
+        }
+    }
+
+    void decorate(GenericBuild genericBuild) {
+        genericBuild.artifacts?.each {
+            decorate(it)
+        }
+    }
+}
--- a/igor-web/src/test/groovy/com/netflix/spinnaker/igor/build/BuildControllerSpec.groovy
+++ b/igor-web/src/test/groovy/com/netflix/spinnaker/igor/build/BuildControllerSpec.groovy
@@ -18,7 +18,6 @@ package com.netflix.spinnaker.igor.build
 
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spinnaker.igor.build.model.GenericBuild
-import com.netflix.spinnaker.igor.jenkins.client.JenkinsClient
 import com.netflix.spinnaker.igor.jenkins.client.model.Build
 import com.netflix.spinnaker.igor.jenkins.client.model.BuildArtifact
 import com.netflix.spinnaker.igor.jenkins.client.model.BuildsList
--- a/igor-web/src/test/groovy/com/netflix/spinnaker/igor/build/artifact/decorator/ConfigurableFileDecoratorSpec.groovy
+++ b/igor-web/src/test/groovy/com/netflix/spinnaker/igor/build/artifact/decorator/ConfigurableFileDecoratorSpec.groovy
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2017 Schibsted ASA.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.igor.build.artifact.decorator
+
+import com.netflix.spinnaker.igor.build.model.GenericArtifact
+import spock.lang.Specification
+import spock.lang.Unroll
+
+class ConfigurableFileDecoratorSpec extends Specification {
+    @Unroll
+    def "decorate artifacts using regex groups"() {
+        when:
+        GenericArtifact genericArtifact = new GenericArtifact(reference, reference, reference)
+        ConfigurableFileDecorator configurableFileDecorator = new ConfigurableFileDecorator(artifactType, decoratorRegex, decoratorRegex)
+        configurableFileDecorator.decorate(genericArtifact)
+
+        then:
+        genericArtifact.name      == name
+        genericArtifact.version   == version
+        genericArtifact.type      == type
+        genericArtifact.reference == reference
+
+        where:
+        artifactType    | decoratorRegex                                 | reference                || name            | version | type
+        "war-release"   | /([a-zA-Z-]+)\-([\d\.]+)\.war$/                | "test-artifact-2.13.war" || "test-artifact" | "2.13"  | "war-release"
+        "jar-release"   | /([a-zA-Z-]+)\-([\d\.]+)\.jar$/                | "test-artifact-2.13.jar" || "test-artifact" | "2.13"  | "jar-release"
+        "Auto resolved" | /([a-zA-Z-]+)\-([\d\.]+)\.([jw]ar)$/           | "test-artifact-2.13.war" || "test-artifact" | "2.13"  | "war"
+        "snapshot-java" |
+            /[\/\:.]*\/([a-zA-Z-]+)\-([\d\.]+\-[\d\.]+)[a-z\-\d]+\.jar$/ |
+            "http://m.a.v/e/n/r/e/p/o/test-artifact/1.3.37-SNAPSHOT/test-artifact-1.3.37-20170304.121217-1337-shaded.jar" ||
+            "test-artifact" |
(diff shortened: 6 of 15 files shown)


gardleopard:
Output without enabling artifact decorator:
```
 ➜ curl -s localhost:8080/builds/status/224/travis-xxx/myorg/igor | jq
{
  "building": false,
  "name": "myorg/igor",
  "number": 224,
  "duration": 68,
  "timestamp": "1479502805000",
  "result": "SUCCESS",
  "artifacts": [
    {
      "fileName": "spinnaker-igor_1.54.0.224.692755-3_all.deb",
      "displayPath": "spinnaker-igor_1.54.0.224.692755-3_all.deb",
      "relativePath": "spinnaker-igor_1.54.0.224.692755-3_all.deb"
    }
  ],
  "url": "https://travis.xx.yyy/myorg/igor/builds/692754",
  "scm": [
    {
      "branch": "sch_master",
      "name": "sch_master",
      "sha1": "efc35b3e781a52f066bda382648fe683c9c9cee6"
    }
  ]
}
```

Output when enabling artifact decorator:
```
curl -s localhost:8080/builds/status/224/travis-xxx/spt-infrastructure/igor | jq
{
  "building": false,
  "name": "myorg/igor",
  "number": 224,
  "duration": 68,
  "timestamp": "1479502805000",
  "result": "SUCCESS",
  "artifacts": [
    {
      "fileName": "spinnaker-igor_1.54.0.224.692755-3_all.deb",
      "displayPath": "spinnaker-igor_1.54.0.224.692755-3_all.deb",
      "relativePath": "spinnaker-igor_1.54.0.224.692755-3_all.deb",
      "reference": "spinnaker-igor_1.54.0.224.692755-3_all.deb",
      "name": "spinnaker-igor",
      "type": "deb",
      "version": "1.54.0.224.692755-3"
    }
  ],
  "url": "https://travis.xx.yyy/myorg/igor/builds/692754",
  "scm": [
    {
      "branch": "sch_master",
      "name": "sch_master",
      "sha1": "efc35b3e781a52f066bda382648fe683c9c9cee6"
    }
  ]
}
```

tomaslin:
I'm still not convinced igor is the service that should have this logic just because it will preclude webhook based events ( such as git triggers ) from being included in the artifact decoration chain. I would suggest putting this in the buildservice in orca since this is always invoked for all events https://github.com/spinnaker/orca/blob/aa808b892abd600d5bbf569793d8b6a8217ec6ed/orca-web/src/main/groovy/com/netflix/spinnaker/orca/controllers/OperationsController.groovy#L121

tomaslin:
another factor to consider is simply that some teams might choose to archive an entire directory in jenkins. Not really a fan of processing hundreds of files to then just filter them out

gardleopard:
# Webhooks:
It will be a challenge with existing webhooks, they need to be changed so that they include the needed metadata. Another thing we could do here is to use igor to decorate the artifacts that are missing decoration. So if a webhook ships an artifact with missing metadata, we can do a call to igor to decorate it. This way we could include the decoration in the chain.

# Orca:
In my opinion orca should know as little as possible about all the logic that lives in spinnaker. We should work hard on limiting the smartness in orca so that its just a service binding the services together and not handling the details. ie https://github.com/spinnaker/orca/blob/fae90b9c751ef31f25de92b52695aaa0b1319a50/orca-bakery/src/main/groovy/com/netflix/spinnaker/orca/bakery/tasks/CompletedBakeTask.groovy#L42

Ideally orca should be a quite stable service that does not need to be updated a lot. The logic should be focused on the bindings, not the details in orca.



# Performance - archive entire directory:
I did some local performance tests on large file lists. 10 000 random artifacts with unknown type, 10 000 rpms and 10 000 debs using this decorator.

The performance looks good, I do not think we less than 100ms to process 10 000 artifacts.

results:
```
10000 unknown extension: 61
10000 deb extension: 71
deb example: name: api     type: deb      version: 1.1.9999-h01.sha123
10000 rpm extension: 49
rpm example: name: openmotif22-libs     type: rpm      version: 2.2.9999
```

Test code:
```
def "performance "() {
        given:
        ArtifactDecorator artifactDecorator = new ArtifactDecorator()
        def benchmark = { closure ->
            def start = System.currentTimeMillis()
            closure.call()
            def now = System.currentTimeMillis()
            now - start
        }
        when:

        GenericBuild genericBuild = new GenericBuild()
        genericBuild.artifacts = new ArrayList<GenericArtifact>()
        10000.times {
            genericBuild.artifacts.add(new GenericArtifact("${it}.${it}", "${it}.${it}", "${it}.${it}"))
        }

        def duration = benchmark {
             artifactDecorator.decorate(genericBuild)

        }
        println("10000 unknown extension: ${duration}")

        genericBuild.artifacts = new ArrayList<GenericArtifact>()
        10000.times {
            genericBuild.artifacts.add(new GenericArtifact("api_1.1.${it}-h01.sha123_all.deb", "api_1.1.${it}-h01.sha123_all.deb", "api_1.1.${it}-h01.sha123_all.deb"))
        }

        duration = benchmark {
            artifactDecorator.decorate(genericBuild)

        }
        println("10000 deb extension: ${duration}")
        println("deb example: ${genericBuild.artifacts.last().prettyPrint()}")

        genericBuild.artifacts = new ArrayList<GenericArtifact>()
        10000.times {
            genericBuild.artifacts.add(new GenericArtifact("openmotif22-libs-2.2.${it}-192.1.3.x86_64.rpm", "openmotif22-libs-2.2.${it}-192.1.3.x86_64.rpm", "openmotif22-libs-2.2.${it}-192.1.3.x86_64.rpm"))
        }
        duration = benchmark {
            artifactDecorator.decorate(genericBuild)

        }
        println("10000 rpm extension: ${duration}")
        println("rpm example: ${genericBuild.artifacts.last().prettyPrint()}")

        then:
        1==1


    }
```

tomaslin:
webhooks get triggered directly in echo, so adding this logic in igor will still miss scenarios like git triggering. Orca already has logic included for handling property files and filtering artifacts. I don't understand the rationale of "making orca stable" by not adding logic there. Putting logic here just means more work and duplication for additional services to have to replicate this artifact logic. 

Let's say we write a jenkins plugin, now the jenkins plugin will have to be able to send the same artifact processing logic as igor. 

Igor is not the right place for this.

gardleopard:
## Git triggering
Git triggers are not affected by this. SCM is something different than artifacts.

## Orca - property files
This does not affect property files in spinnaker, the property file handling is not touched in this PR.

## Orca - filtering artifacts
Orca will still be the master of filtering artifacts, but it should filter decorated artifacts, using type and name to find what is relevant, not knowing the details about artifacts.

## Orca - making orca stable
What I wrote above was "Ideally orca should be a quite stable service that does not need to be updated a lot. The logic should be focused on the bindings, not the details in orca." By this I mean that Orca should be loosely coupled to the rest of Spinnaker. Ideally we should be able to add new cloud providers without even touching orca.

## Duplication of artifact logic
If you want to have a jenkins plugin that pushes webhooks to echo, it needs to provide, type, name, version and reference in the artifact data. The current way of working with artifacts in spinnaker is using the raw jenkins representation, and this representation is created to represent files made available by jenkins. Spinnaker should have a representation of artifacts that makes artifacts easy to handle in spinnaker, not stick with what was there first.

## Igor is not the right place for this
Only the producers of artifacts in Spinnaker should know the details about the artifacts, so that the knowledge is encapsulated where is belongs and not distributed across Spinnaker. If we do this as early as possible, its easier, because we have the details where we produce the artifact.

## More
This discussion is focused around whats already in spinnaker, and we still will support whats there. The big picture in this is what we can do after artifacts is decorated and what we can gain by this change. For instance lets say that a team uses boxfuse to bake amis fast. With this feature they could easily create a gradle plugin or whatever to publish events like:
```artifacts [
{
"type":"ami",
"name":"boxfuse-created-ami",
"version":"1.3.3.7",
"reference":"ami-41ca"
}
]
```
Currently this can be done by adding this reference to a property file and do some magic in the pipeline. If we use decorated artifacts, we do not need any magic. We select "boxfuse-created-ami" in the deployment stage and this is resolved from the trigger context.

duftler:
Discussed this with @tomaslin offline, and we'll go with this approach for now.

Since these changes will be wide-ranging, and will touch several services, please create an issue on https://github.com/spinnaker/spinnaker/issues and associate all the relevant PR's with that issue.

Lastly, it would be helpful to add to this PR description a summary of some of the items we discussed offline. For example:
- Existing behavior will be preserved
- Eventually, a producer will be able to submit artifacts that are either decorated or undecorated
- Decorated artifacts will be propagated without being modified
- Echo can call igor to process undecorated artifacts

And so on...

@lwander please review this PR when you have time.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/service/ArtifactDecorator.groovy
```diff
@@ -0,0 +1,62 @@
...
+        artifactDetailsDecorators.put("deb", new DebDetailsDecorator())
+        artifactTypeIdentifiers.add(new FileArtifactIdentifier())
+    }
+
+    void decorate(GenericArtifact genericArtifact) {
+        String type = getType(genericArtifact.fileName)
+        ArtifactDetailsDecorator artifactDetailsDecorator = artifactDetailsDecorators.get(type)
+        if(artifactDetailsDecorator) {
+            artifactDetailsDecorator.decorate(genericArtifact)
+
```
lwander: Extra newline

igor-web/src/main/groovy/com/netflix/spinnaker/igor/service/ArtifactDecorator.groovy
```diff
@@ -0,0 +1,62 @@
...
+import com.netflix.spinnaker.igor.build.artifact.decorator.DebDetailsDecorator
+import com.netflix.spinnaker.igor.build.artifact.decorator.RpmDetailsDecorator
+import com.netflix.spinnaker.igor.build.artifact.identifier.ArtifactTypeIdentifier
+import com.netflix.spinnaker.igor.build.artifact.identifier.FileArtifactIdentifier
+import com.netflix.spinnaker.igor.build.model.GenericArtifact
+import com.netflix.spinnaker.igor.build.model.GenericBuild
+
+class ArtifactDecorator {
+
+    Map<String, ArtifactDetailsDecorator> artifactDetailsDecorators = new HashMap<String, ArtifactDetailsDecorator>()
```
lwander: This might be controversial, but it feels like these should be autowired to keep in spirit with the rest of Igor.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/service/ArtifactDecorator.groovy
```diff
@@ -0,0 +1,62 @@
...
+    ArtifactDecorator() {
+        artifactDetailsDecorators.put("rpm", new RpmDetailsDecorator())
+        artifactDetailsDecorators.put("deb", new DebDetailsDecorator())
+        artifactTypeIdentifiers.add(new FileArtifactIdentifier())
+    }
+
+    void decorate(GenericArtifact genericArtifact) {
+        String type = getType(genericArtifact.fileName)
+        ArtifactDetailsDecorator artifactDetailsDecorator = artifactDetailsDecorators.get(type)
+        if(artifactDetailsDecorator) {
```
lwander: Missing space before `(`

igor-web/src/main/groovy/com/netflix/spinnaker/igor/service/ArtifactDecorator.groovy
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    void decorate(GenericBuild genericBuild) {
+        genericBuild.artifacts?.each {
+            decorate(it)
+        }
+    }
+
+    String getType( String reference ) {
+        for(ArtifactTypeIdentifier identifier: artifactTypeIdentifiers) {
```
lwander: Missing space before `(`

igor-web/src/main/groovy/com/netflix/spinnaker/igor/service/ArtifactDecorator.groovy
```diff
@@ -0,0 +1,62 @@
...
+            String artifactType = identifier.artifactType(reference)
+            if(artifactType) {
```
lwander: Missing space before `(`

igor-web/src/main/groovy/com/netflix/spinnaker/igor/service/ArtifactDecorator.groovy
```diff
@@ -0,0 +1,62 @@
...
+        }
+    }
+
+    void decorate(GenericBuild genericBuild) {
+        genericBuild.artifacts?.each {
+            decorate(it)
+        }
+    }
+
+    String getType( String reference ) {
```
lwander: How do you know that you can always get the type from a reference? What if the reference isn't enough?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/service/ArtifactDecorator.groovy
```diff
@@ -0,0 +1,62 @@
...
+
+    ArtifactDecorator() {
+        artifactDetailsDecorators.put("rpm", new RpmDetailsDecorator())
+        artifactDetailsDecorators.put("deb", new DebDetailsDecorator())
+        artifactTypeIdentifiers.add(new FileArtifactIdentifier())
+    }
+
+    void decorate(GenericArtifact genericArtifact) {
+        String type = getType(genericArtifact.fileName)
+        ArtifactDetailsDecorator artifactDetailsDecorator = artifactDetailsDecorators.get(type)
```
lwander: This relationship between types and decorators should be made stronger. If I were to implement a new decorator, I shouldn't have to update a hashmap, implement a new type identifier (and make sure no other identifiers misidentify my type), as well as implement a separate decorator. This implementation path leaves room for error, that could be avoided by providing a stronger coupling between types and decorators.

gardleopard:
This is the first PR for https://github.com/spinnaker/spinnaker/issues/1348

igor-web/src/main/groovy/com/netflix/spinnaker/igor/service/ArtifactDecorator.groovy
robzienert: This is a bit meta, but do we have any thoughts on adding something like codenarc so we can fail builds when PRs do not conform to style conventions?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/artifact/identifier/ArtifactTypeIdentifier.groovy
```diff
@@ -0,0 +1,23 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.igor.build.artifact.identifier
+
+interface ArtifactTypeIdentifier {
```
gardleopard: Good idea, I will give it a go.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/service/ArtifactDecorator.groovy
gardleopard: Thanks, I will investigate this.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/model/GenericArtifact.groovy
```diff
@@ -16,10 +16,17 @@
 
 package com.netflix.spinnaker.igor.build.model
 
+import com.fasterxml.jackson.annotation.JsonInclude
+
+@JsonInclude(JsonInclude.Include.NON_NULL)
 class GenericArtifact {
     String fileName
     String displayPath
     String relativePath
```
gardleopard: Im not sure, when we roll this out they will stay at least, when artifact decoration is used throughout spinnaker we can deprecate and remove them.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/artifact/decorator/RpmDetailsDecorator.groovy
```diff
@@ -0,0 +1,60 @@
...
+        genericArtifact.version = extractVersion(genericArtifact.fileName)
+        genericArtifact.name = extractName(genericArtifact.fileName)
+        genericArtifact.type = packageType
+        genericArtifact.reference = genericArtifact.fileName
+        return genericArtifact
+    }
+
+    @Override
+    boolean knowsThisArtifact(GenericArtifact genericArtifact) {
+        //TODO FRAGILE
```
lwander: Should this same todo live in the DebDetailsDecorator as well?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/service/ArtifactDecorator.groovy
```diff
@@ -0,0 +1,68 @@
...
+    List<ArtifactDetailsDecorator> artifactDetailsDecorators;
+
+    @Autowired(required = false)
+    ArtifactDecorationProperties artifactDecorationProperties
+
+    @Autowired
+    ArtifactDecorator(List<ArtifactDetailsDecorator> artifactDetailsDecorators, @Valid ArtifactDecorationProperties artifactDecorationProperties) {
+        this.artifactDetailsDecorators = artifactDetailsDecorators
+        List<ArtifactDetailsDecorator> configuredArtifactDetailsDecorators = artifactDecorationProperties?.fileDecorators?.collect { ArtifactDecorationProperties.FileDecorator fileDecorator ->
+            log.error "Configuring type : ${fileDecorator.type}"
```
lwander: Why `log.error`?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/artifact/decorator/RpmDetailsDecorator.groovy
gardleopard: I have created tests that verifies the this actually is good enough, so ill remove the TODO.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/service/ArtifactDecorator.groovy
gardleopard: Logg error is wrong, i will change it to info for this.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/model/GenericArtifact.groovy
```diff
@@ -16,14 +16,49 @@
...
+    String type
+    String version
 
     GenericArtifact(String fileName, String displayPath, String relativePath) {
         this.fileName = fileName
         this.displayPath = displayPath
         this.relativePath = relativePath
     }
+
+    String toString() {
```
jervi: This will not produce the output you want:
`fileName:,somefilename.jardisplayPath:,somedisplaypathrelativePath:,somerelativepathreference:,This,is,a,referencename:,Name,of,artifacttype:,jarversino:,123`
You could instead just annotate the class with `@ToString(includeNames=true)` and remove this method.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/artifact/decorator/ConfigurableFileDecorator.groovy
```diff
@@ -0,0 +1,52 @@
...
+
+import com.netflix.spinnaker.igor.build.model.GenericArtifact
+import groovy.util.logging.Slf4j
+
+@Slf4j
+class ConfigurableFileDecorator implements ArtifactDetailsDecorator {
+
+    String decoratorRegex
+    String identifierRegex
+    String type
```
jervi: These fields could be final

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/model/GenericArtifact.groovy
gardleopard: Thanks @jervi, I have made the change now and it is a lot better!

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/model/GenericArtifact.groovy
```diff
@@ -16,14 +16,49 @@
...
+            sb.append("reference: ${reference}")
+        }
+        if (name) {
+            sb.append("name: ${name}")
+        }
+        if (type) {
+            sb.append("type: ${type}")
+        }
+        if (version) {
+            sb.append("versino: ${version}")
```
gardleopard: removed

duftler:
Please update the description on the commit to reflects the comments in my note on Dec 13, 2016. If we don't update the top-level description, the log message for this commit may be misleading/incomplete.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/service/ArtifactDecorator.groovy
```diff
@@ -0,0 +1,69 @@
...
+            (ArtifactDetailsDecorator) new ConfigurableFileDecorator(fileDecorator.type, fileDecorator.decoratorRegex, fileDecorator.identifierRegex)
+        }
+        if (configuredArtifactDetailsDecorators) {
+            this.artifactDetailsDecorators.addAll(0, configuredArtifactDetailsDecorators)
+        }
+        this.artifactDetailsDecorators
+    }
+
+    void decorate(GenericArtifact genericArtifact) {
+        for (ArtifactDetailsDecorator artifactDetailsDecorator : artifactDetailsDecorators) {
```
duftler: This routine could be written like:
```
        ArtifactDetailsDecorator artifactDetailsDecorator = artifactDetailsDecorators.find { it.handles(genericArtifact) }

        if (artifactDetailsDecorator) {
            artifactDetailsDecorator.decorate(genericArtifact)
            log.debug "Decorated artifact: $genericArtifact"
        }
```

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/artifact/decorator/RpmDetailsDecorator.groovy
```diff
@@ -0,0 +1,59 @@
...
+@Component
+@ConditionalOnProperty('artifact.decorator.enabled')
+class RpmDetailsDecorator implements ArtifactDetailsDecorator {
+
+    String packageType = 'rpm'
+    String versionDelimiter = '-'
+
+    @Override
+    GenericArtifact decorate(GenericArtifact genericArtifact) {
+        genericArtifact.version = extractVersion(genericArtifact.fileName)
```
duftler: The order is `name`, `version` in `DebDetailsDecorator` just above.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/build/BuildControllerSpec.groovy
```diff
@@ -27,6 +27,7 @@ import com.netflix.spinnaker.igor.jenkins.client.model.ParameterDefinition
 import com.netflix.spinnaker.igor.jenkins.client.model.QueuedJob
 import com.netflix.spinnaker.igor.jenkins.service.JenkinsService
 import com.netflix.spinnaker.igor.model.BuildServiceProvider
+import com.netflix.spinnaker.igor.service.ArtifactDecorator
```
duftler: I don't think this import is needed.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/build/BuildControllerSpec.groovy
```diff
@@ -81,7 +82,7 @@ class BuildControllerSpec extends Specification {
         buildMasters = Mock(BuildMasters)
         server = new MockWebServer()
         mockMvc = MockMvcBuilders.standaloneSetup(new BuildController(
-            executor: Executors.newSingleThreadExecutor(), buildMasters: buildMasters, objectMapper: new ObjectMapper())).build()
+            executor: Executors.newSingleThreadExecutor(), buildMasters: buildMasters, objectMapper: new ObjectMapper(), artifactDecorator: null)).build()
```
duftler: Won't `artifactDecorator` already be null?

igor-web/src/test/groovy/com/netflix/spinnaker/igor/build/artifact/decorator/ConfigurableFileDecoratorSpec.groovy
```diff
@@ -0,0 +1,66 @@
...
+        ConfigurableFileDecorator configurableFileDecorator = new ConfigurableFileDecorator(artifactType, decoratorRegex, decoratorRegex)
+        configurableFileDecorator.decorate(genericArtifact)
+
+        then:
+        genericArtifact.name    == name
+        genericArtifact.version == version
+        genericArtifact.type    == type
+
+        where:
+        artifactType    || decoratorRegex                                 || reference                || name            || version || type
```
duftler: Typically `|` is for all columns up until the expected results.
So something like:
`given1 | given2 | given3 || expectedResult`

duftler: Here should probably be:
`artifactType    | decoratorRegex                                 | reference                || name            | version | type`

igor-web/src/test/groovy/com/netflix/spinnaker/igor/build/artifact/decorator/DebDetailsDecoratorSpec.groovy
```diff
@@ -0,0 +1,90 @@
...
+package com.netflix.spinnaker.igor.build.artifact.decorator
+
+import com.netflix.spinnaker.igor.build.model.GenericArtifact
+import spock.lang.Specification
+import spock.lang.Unroll
+
+class DebDetailsDecoratorSpec extends Specification {
+
+    @Unroll
+    def "set version on DEBs"() {
```
duftler: Maybe "identifies version from DEBs". Or not, don't feel too strongly about it.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/service/ArtifactDecoratorSpec.groovy
```diff
@@ -0,0 +1,64 @@
...
+        GenericArtifact genericArtifact = new GenericArtifact(reference, reference, reference)
+        artifactDecorator.decorate(genericArtifact)
+
+        then:
+        genericArtifact.name    == name
+        genericArtifact.version == version
+        genericArtifact.type    == type
+
+        where:
+        reference || name || version || type
```
duftler: Please format this table such that the columns are aligned.

gardleopard:
@duftler I have addressed all the comments now. Thanks for reviewing.

gardleopard:
I found a bug that I did not set reference for the configurable decorators. I have added code for that in a separate commit here.

PR #140, opened 2016-12-01
Parse a full JSON line from travis logs based on a magic string

PerGon:
This PR allows to extract multiple properties from the Travis log. The properties are printed in one line in the Travis log in JSON format and prepended by `SPINNAKER_CONFIG_JSON=`.
Something like this:
```
SPINNAKER_CONFIG_JSON={"key1": "value1", "key2": "value2", "key3": {"key3_1": "value3_1", "key3_2": "value3_2"} }
```

As a side note: I'm swallowing the exception in case the JSON is malformed - doing it in a similar fashion as the [Jenkins implementation](https://github.com/spinnaker/igor/blob/master/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy#L224). Possibily the exception should be handled differently, but that means changing Orca as well - shall be addressed in a different PR.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/client/logparser/PropertyParser.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/client/logparser/PropertyParser.groovy
@@ -16,13 +16,17 @@
 
 package com.netflix.spinnaker.igor.travis.client.logparser
 
+import com.fasterxml.jackson.databind.ObjectMapper
 import groovy.transform.CompileStatic
 import groovy.util.logging.Slf4j
 
 @CompileStatic
 @Slf4j
 class PropertyParser {
+
     static final String MAGIC_SEARCH_STRING = "SPINNAKER_PROPERTY_"
+    static final String MAGIC_JSON_SEARCH_STRING = "SPINNAKER_CONFIG_JSON="
+
     static Map<String, Object> extractPropertiesFromLog(String buildLog) {
         Map<String, Object> map = new HashMap<String, Object>()
         buildLog.split('\n').each { String line ->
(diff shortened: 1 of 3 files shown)


gardleopard:
@spinnaker/reviewers PTAL

igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/client/logparser/PropertyParser.groovy
```diff
@@ -34,6 +38,17 @@ class PropertyParser {
                 log.info "${key}:${value}"
                 map.put(key,value)
             }
+            if (line =~ MAGIC_JSON_SEARCH_STRING) {
```
lwander: Looks like you intend to have the magic string be the first non-whitespace substring in the line, would it be worth matching on that? `^\s*${MAGIC_JSON....`

lwander:
Hey @gardleopard how does this fit in with your artifact refactor - is the goal to replace this with a proper json artifact? Or will this be the only way to supply more than one artifact to a pipeline? (json in the logs + artifact produced by build).

gardleopard:
@lwander I think this fits in, the properties have a slightly different use case. They open up for customisations that would be hard without it. And this PR makes a lot better in travis, because you can supply a complete json instead of one and one field as the solution currently is.

This will not be the only way to supply more than one artifact to a pipeline, but rather to supply different properties that are useful to the pipeline.

gardleopard:
@tomaslin Please take a look on this, the PR here will make it a lot easier for us to use properties from travis jobs in our pipelines.

tomaslin:
LGTM

PerGon:
Thank you all for your comments.

@tomaslin branch is now rebased and up-to-date with master.

PR #142, opened 2016-12-16
feat(config/travis): configurable regexes to parse artifacts

srvaroa:
travis: configurable regexes to parse artifacts

Offer configurable regexes to find RPM/DEB artifacts pushed to
artifactory as the default ones only match on the output of the "art"
CLI, but wouldn't on that of any other tools.  The result is a failed
Spinnaker pipeline when no artifact is found.

With this change users may add a list of arbitrary regexes that will be
used to parse build logs.  An example is provided that corresponds to
Gradle maven-publish, which would match on the following line (taken
from a real build, with project names changed):

        Upload https://artifacts.my.org/artifactory/yum-local/some/project/package-0.1.0.20161216.g1b1c767.rpm

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,15 @@ travis:
     baseUrl: https://travis-ci.org
     address: https://api.travis-ci.org
     githubToken: 6a7729bdba8c4f9abc58b175213d83f072d1d832
+  regexes:
+  - /Upload https?:\/\/.+\/(.+\.(deb|rpm))/
 ```
 
 Currently Travis is used within Spinnaker to trigger pipelines and provide artifact information for the bake stages.
+
+When parsing artifact information from Travis builds, Igor uses a default regex
+that will match on output from the `art` CLI tool.  Different regexes than the
+default may be configured using the `regexes` list.
 
 ## Git Repositories
 
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/client/logparser/ArtifactParser.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/client/logparser/ArtifactParser.groovy
@@ -20,22 +20,34 @@ import com.netflix.spinnaker.igor.build.model.GenericArtifact
 import groovy.transform.CompileStatic
 import groovy.util.logging.Slf4j
 
+import javax.xml.bind.annotation.adapters.XmlJavaTypeAdapter
+import java.util.regex.Matcher
+
 @CompileStatic
 @Slf4j
 class ArtifactParser {
 
-    static def REGEXES = [ /Uploading artifact: https?:\/\/.+\/(.+\.(deb|rpm)).*/,
-                          /Successfully pushed (.+\.(deb|rpm)) to .*/ ]
+    static List<String> DEFAULT_REGEXES = [
+        /Uploading artifact: https?:\/\/.+\/(.+\.(deb|rpm)).*$/,
+        /Successfully pushed (.+\\.(deb|rpm)) to .*/].toList()
 
-    static List<GenericArtifact> getArtifactsFromLog(String buildLog) {
-        List<GenericArtifact> artifacts = new ArrayList<GenericArtifact>()
+    /**
+     * Parse the build log using the given regular expressions.  If they are
+     * null, or empty, then DEFAULT_REGEXES will be used, matching on artifacts
+     * uploading from the `art` CLI tool.
+     */
+    static List<GenericArtifact> getArtifactsFromLog(String buildLog, Iterable<String> regexes) {
+        final List<GenericArtifact> artifacts = new ArrayList<GenericArtifact>()
+        if (regexes == null || regexes.size() <= 0) {
+            regexes = DEFAULT_REGEXES
+        }
         buildLog.split('\n').each { line ->
-            def m
-            for (def regex : REGEXES) {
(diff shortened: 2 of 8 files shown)


srvaroa:
Sorry for the multiple pushes, I am unable to build locally due to:

* What went wrong:
A problem occurred evaluating root project 'igor'.
> Failed to apply plugin [class 'com.netflix.spinnaker.gradle.project.SpinnakerProjectConventionsPlugin']
   > index is out of range 0..-1 (index = 0)

srvaroa:
@tomaslin @gardleopard let me know if this needs any improvements, thanks!

gardleopard:
I would like it to keep the one there as the default and have a configuration override for adding custom parsers. ie some configurable parser setting for each travis server in https://github.com/spinnaker/igor/blob/master/igor-web/config/igor.yml

srvaroa:
Thanks @gardleopard, that makes sense. I have this in my queue and will refactor as you suggest as soon as possible.

srvaroa:
@gardleopard @tomaslin let me know how this looks now

igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/client/logparser/ArtifactParser.groovy
```diff
@@ -24,18 +24,31 @@ import groovy.util.logging.Slf4j
...
+        return getArtifactsFromLog(buildLog, Collections.EMPTY_LIST)
+    }
+
+    /**
+     * Parse the build log using DEFAULT_REGEXES plus the given ones.
+     */
+    static List<GenericArtifact> getArtifactsFromLog(String buildLog, Iterable<String> additionalRegexes) {
+        final List<GenericArtifact> artifacts = new ArrayList<GenericArtifact>()
+        final Set<String> regexes = new HashSet<>()
+        regexes.addAll(DEFAULT_REGEXES)
```
gardleopard: I would prefer if injected regexes replaces the default ones. In some cases you would want to not use the standard regexes.

README.md
```diff
@@ -1,87 +1,96 @@
...
-dockerRegistry:
-  enabled: true
-```
-
-## Running Igor
-
-Igor requires redis server to be up and running.
-
-Start Igor via `./gradlew bootRun`. Or by following the instructions using the [Spinnaker installation scripts](https://www.github.com/spinnaker/spinnaker).
+# Igor
```
gardleopard: Why is all this text replaced? You have not changed it all.

README.md
```diff
@@ -1,87 +1,96 @@
...
+```
+
+Currently Travis is used within Spinnaker to trigger pipelines and provide artifact information for the bake stages.
+
+When parsing artifact information from Travis builds, Igor uses a default regex
+that will match on output from the `art` CLI tool.  Other regexes may be added
+by using a `regexes` list:
+
+```
+# regexes:
```
gardleopard: I would move this up to travis so the placement for this is clear.

README.md
srvaroa: I didn't realize the file has DOS line endings and I have ff=UNIX in my vimrc.  I'll respect the current settings.

srvaroa:
@gardleopard I had the example already at [igor-web/config/igor.yml](https://github.com/spinnaker/igor/pull/142/commits/81858690198f94771066ec1c6a0027f53134c9b0#diff-9697b65a9afb6f42b188fdfead112d97R24) but let me know if I'm missing the right place.

tomaslin:
is there any updates on this issue @gardleopard ?

gardleopard:
Needs tests for parsing the yaml configuration, could not get it to work in our environment.

srvaroa:
Thanks @gardleopard, I will add some extra checks.

srvaroa:
@gardleopard I have been starting the service with gradle run and different values in igor.yml.  I think the mistake was just in the sample values provided, Groovy accepts / when declaring string literals so they are fine in the test.  But in the yaml they are taken as part of the string. Just removing the surrounding / gets the regexes correctly loaded into the TravisService.

I updated this in igor.yml and made minor changes to the PR, but it's mostly as-is.  Could you try with any of these config values to verify?

    - Uploading artifact: https?:\/\/.+\/(.+\.(deb|rpm)).*
    - Successfully pushed (.+\.(deb|rpm)) to .*
    - Uploading artifact:\s+.*?([^\/]+\.(deb|rpm)).*

(Unrelated to this PR, but I noticed that the sample values for `decoratorRegex` and `identifierRegex` in igor.yml may have the same problem.)

gardleopard:
This is got to go for me.
The examples above do not actually work it needs to be
```
    - Successfully pushed (.+\.(deb|rpm)) to .*
    - 'Uploading artifact:\s+.*?([^\/]+\.(deb|rpm)).*'
```
The `:` is interpreted by yaml and has to be quoted. I think as long as you are quoting the regex this will just work.


@tomaslin This is PR is good to be merged in my opinion.

tomaslin:
LGTM

tomaslin:
Can you change the commit message so it fits within the spinnaker commit format? http://www.spinnaker.io/docs/how-to-submit-a-patch#section-commit-message-conventions

srvaroa:
Thanks @tomaslin , @gardleopard - I've updated the PR & commit, let me know if I missed any detail.

PR #145, opened 2016-12-28
Only publish events to echo if a build has been seen at least once

ajordens:
This offers some protection around the loss (or flush) of an underlying redis store.

Currently, igor will publish events for _all_ jenkins builds (regardless of completion date) if redis is flushed.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
@@ -198,8 +198,7 @@ class JenkinsBuildMonitor implements PollingMonitor {
                                                         Project oldProject = new Project(name: project.name, lastBuild: build)
                                                         if (build.number != lastBuild
                                                                 || (build.number == lastBuild && cachedBuild.lastBuildBuilding != build.building)) {
-                                                            echoService.postEvent(
-                                                                    new BuildEvent(content: new BuildContent(project: oldProject, master: master)))
+                                                            postEvent(echoService, cachedBuilds, oldProject, master)
                                                         }
                                                     } catch (e) {
                                                         log.error("An error occurred fetching ${master}:${project.name}:${build.number}", e)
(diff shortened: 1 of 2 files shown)


ajordens:
This does nothing for re-triggering of docker tags in the event of a redis flush, but does attempt to offer some protection for jenkins in the event that the underlying redis instance is terminated or flushed.

Seemed simple enough to filter in `igor`. A terrible idea @tomaslin ?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
```diff
@@ -198,8 +198,7 @@ class JenkinsBuildMonitor implements PollingMonitor {
                                                         Project oldProject = new Project(name: project.name, lastBuild: build)
                                                         if (build.number != lastBuild
                                                                 || (build.number == lastBuild && cachedBuild.lastBuildBuilding != build.building)) {
-                                                            echoService.postEvent(
-                                                                    new BuildEvent(content: new BuildContent(project: oldProject, master: master)))
+                                                            postEvent(echoService, cachedBuild, oldProject, master)
```
tomaslin: I'm not sure this handles new jobs correctly. If I have a new jenkins job with completed runs when the index runs, say build 1 and build 2, it would look like the events for those builds will get swallowed and not triggered again. Seeing that adding new jobs is more likely than losing redises, I'm not sure this is the best way to do this. Can we not just check the number of indexed builds per master and only do this when that particular master has 0 indexed builds?

ajordens: Yes, it would trigger subsequent builds if the redis is lost ... I figured it unlikely someone would add a new job and setup a trigger before we indexed the job.

One could argue that the impact of a lost redis is more severe than failing to trigger on the first build of a brand new job ... even if it's something that seldom occurs.

I don't even know if we need to check indexed builds per master ... might be able to get away with just checking that anything has been indexed (ie. `cachedBuilds.isEmpty()`) ... I'll play around with that.

ajordens: We're on the same page ... what I'm calling `cachedBuilds` is actually the indexed builds for the master.

I've updated the PR to check that.

PR #153, opened 2017-02-20
Add BitBucket Cloud support

robumberger:
PR's are being opened against Deck, Echo, and Gate.

BitBucket Server (a.k.a. Stash) support already exists. This adds BitBucket Cloud support of pipeline triggers based on BitBucket webhooks and provides Orca the ability to retrieve a list of commits from BitBucket as part of it's GetCommits stage.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/ScmInfoController.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/ScmInfoController.groovy
@@ -18,6 +18,7 @@ package com.netflix.spinnaker.igor.scm
 
 import com.netflix.spinnaker.igor.scm.github.client.GitHubMaster
 import com.netflix.spinnaker.igor.scm.stash.client.StashMaster
+import com.netflix.spinnaker.igor.scm.bitbucket.client.BitBucketMaster
 import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.web.bind.annotation.RequestMapping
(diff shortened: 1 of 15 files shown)


robumberger:
Associated PR's:

Deck: [#3319](https://github.com/spinnaker/deck/pull/3319)
Echo: [#137](https://github.com/spinnaker/echo/pull/137)
Gate: [#346](https://github.com/spinnaker/gate/pull/346)

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/ScmInfoController.groovy
```diff
@@ -37,6 +38,9 @@ class ScmInfoController {
     @Autowired(required = false)
     GitHubMaster gitHubMaster
 
+    @Autowired(required = false)
+    BitBucketMaster bitBucketMaster
```
tomaslin: This should be consistent with the stash support and have this under /bitbucket/ commit controller instead of joining github and bitbucket

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/ScmInfoController.groovy
```diff
@@ -46,6 +50,9 @@ class ScmInfoController {
 
         if(gitHubMaster)
             result << [gitHub : gitHubMaster.baseUrl]
+
+        if(bitBucketMaster)
```
tomaslin: I am not sure this change is needed. It looks like we have a list of github masters but bitbucket cloud api only allows one endpoint.

robumberger: Correct, there is only a single endpoint. This was added to mimic the Stash implementation (see line 43/47).

tomaslin: That is for githubMaster...

tomaslin: Ah, I see. Nvm.

PR #160, opened 2017-05-02
feat(travis/pro) Support fetching of logs from travis pro

gardleopard:
The travis pro api does not have log_id in the api response for jobs. This means we need to find the logs slightly different.
Travis has three products Open source, Pro and Enterprise, they all behave a little bit different.

@abrahamoshel can you test that this works for you?


abrahamoshel:
It works. Fired off the trigger as expected and has all the relevant information for the rest of the pipeline to work. Thanks @gardleopard

gardleopard:
@spinnaker/reviewers PTAL

tomaslin:
LGTM

PR #161, opened 2017-05-03
fix(jenkins): Fixing missed build events on concurrent builds

jeyrschabu:
- Removed the use of parallelStream while processing builds
- parallelStream may lead to a concurrency issue with redis read/write for close enough builds
- Refactored method for readability


jeyrschabu:

tomaslin:
LGTM didn't test functionality

PR #165, opened 2017-05-13
fix(jenkins): Changing jenkins monitor algorithm to process builds using timestamp ranges

jeyrschabu:
- Uses a job's last build timestamp and sets the lower bound for the interval in Redis
- Uses a sliding upperbound as new builds appear on subsequent polls
- Ensures only one event gets fired for a completed build

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
@@ -22,7 +22,9 @@ import com.netflix.spinnaker.igor.IgorConfigurationProperties
 import com.netflix.spinnaker.igor.history.EchoService
 import com.netflix.spinnaker.igor.history.model.BuildContent
 import com.netflix.spinnaker.igor.history.model.BuildEvent
+import com.netflix.spinnaker.igor.jenkins.client.model.Build
 import com.netflix.spinnaker.igor.jenkins.client.model.Project
+import com.netflix.spinnaker.igor.jenkins.service.JenkinsService
 import com.netflix.spinnaker.igor.model.BuildServiceProvider
 import com.netflix.spinnaker.igor.polling.PollingMonitor
 import com.netflix.spinnaker.igor.service.BuildMasters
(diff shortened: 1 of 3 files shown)


jeyrschabu:
The current logic makes an incorrect assumption that builds will finish sequentially by number. For example: builds #1, #2, #3 will finish in that order. This logic falls apart when multiple builds are launched and finish randomly. This causes us to miss triggers on these builds.

The new logic uses timestamps for ranges [lastPollstamp - job.lastBuild.timestamp). By using a sliding upper bound, we ensure to account for new builds as they come in. We advance the lower bound cursor only when all builds have completed.

jeyrschabu:

igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
```diff
@@ -134,123 +133,87 @@ class JenkinsBuildMonitor implements PollingMonitor {
...
+                if (!job.lastBuild) {
+                    log.info("no builds on ${master}:${job.name} skipping...")
+                    continue
+                }
+
+                Long cursor = cache.getLastPollCycleTimestamp(master, job.name)
+                Long lastBuildStamp = job.lastBuild.timestamp as Long
+                Date upperBound = new Date(lastBuildStamp)
+
+                log.info("last cursor was ${cursor}, last build on this job was at ${upperBound}")
```
tomaslin: The logs for this will be really noisy, can we clean it up so it logs the job name & build number only when there is a hit?

tomaslin: also log should have master and job.name at least

igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
```diff
@@ -134,123 +133,87 @@ class JenkinsBuildMonitor implements PollingMonitor {
...
+                            log.info("posting build event for build ${master}:${job.name}:${build.number}")
+                            postEvent(echoService, new Project(name: job.name, lastBuild: build), master)
+                            cache.setEventPosted(master, job.name, cursor, build.number)
                         }
                     }
-            )
-
-            log.info("Took ${System.currentTimeMillis() - startTime}ms to retrieve projects (master: ${master})")
 
+                    // 3. advance cursor when all builds have completed in the interval
```
tomaslin: It feels like this logic could be simpler if you set the cursor to be the highest timestamp across all builds instead of the timestamp of the last build. This way you don't have to keep track of builds that are still in progress and can look at only completed builds.

jeyrschabu: Addressed this in discussion

PR #168, opened 2017-06-12
feat(igor/travis) Filter builds on branches, tags and pull requests to…

gardleopard:
… populate reasonable amount of builds when manually triggering pipelines

The old implementation of this only took the last 25 builds from the travis job and filtered that list to give the user a list of builds to select from. Now we filter as much as we can in the request and this gives the user a more consistent experience when selecting builds to trigger with.

Tags is still handled differently, because there is no filter in the travis api to get only tag builds.


gardleopard:
@spinnaker/reviewers PTAL

lwander:
Ready to merge?

PR #177, opened 2017-07-24
feat(jenkins) allow polling to be turned off

kevinawoo:
Adding the ability to turn off jenkins polling. Its now possible to run multiple igors without duplicate notifications to echo.

Best viewed without whitespace changes (however, line commenting is not possible)
https://github.com/armory-io/igor/pull/1/files?w=1 

#### Key take aways
- polling is on by default
- igor's api still servers the jenkins info
- no jenkins events are sent to echo
- any changes in jenkins does not persist to cache
- there must be at least 1 polling igor to kick off jobs.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
@@ -30,6 +30,7 @@ import com.netflix.spinnaker.igor.polling.PollingMonitor
 import com.netflix.spinnaker.igor.service.BuildMasters
 import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.beans.factory.annotation.Value
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.context.event.ContextRefreshedEvent
 import org.springframework.core.env.Environment
@@ -53,6 +54,9 @@ class JenkinsBuildMonitor implements PollingMonitor {
 
     @Autowired
     Environment environment
+
+    @Value('${jenkins.polling.enabled:true}')
+    boolean pollingEnabled = true
 
     Scheduler scheduler = Schedulers.io()
     Worker worker = scheduler.createWorker()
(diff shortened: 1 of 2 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
```diff
@@ -108,21 +112,23 @@ class JenkinsBuildMonitor implements PollingMonitor {
...
-                                { master -> changedBuilds(master) }
-                        )
-                        log.info "- Polling cycle done - ${new Date()}"
-                    } else {
-                        log.info("not in service (lastPoll: ${lastPoll ?: 'n/a'})")
-                        lastPoll = null
-                    }
-                } as Action0, 0, pollInterval, TimeUnit.SECONDS
-        )
+        if (jenkinsProperties.polling.enabled) {
```
tomaslin: Would use a Spring @Value with a default of true instead of wiring this in jenkinsProperties and  bind that to the inService() method instead of current implementation

kevinawoo:
@tomaslin thanks! Its much cleaner now 😀

igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
```diff
@@ -95,14 +99,14 @@ class JenkinsBuildMonitor implements PollingMonitor {
     boolean isInService() {
         if (discoveryClient == null) {
             log.info("no DiscoveryClient, assuming InService")
-            true
+            pollingEnabled
```
tomaslin: for readability this should be `return pollingEnabled` since the return value is happening in the middle of the method

igor-web/src/test/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitorSchedulingSpec.groovy
```diff
@@ -85,4 +85,42 @@ class JenkinsBuildMonitorSchedulingSpec extends Specification {
...
+        monitor = new JenkinsBuildMonitor(cache: cache, buildMasters: buildMasters, igorConfigurationProperties: cfg)
+        monitor.pollingEnabled = false
+        monitor.worker = scheduler.createWorker()
+
+        when:
+        monitor.onApplicationEvent(Mock(ContextRefreshedEvent))
+        scheduler.advanceTimeBy(1L, TimeUnit.SECONDS.MILLISECONDS)
+
+        then: 'initial poll'
+        0 * buildMasters.filteredMap(BuildServiceProvider.JENKINS) >> [MASTER: jenkinsService]
```
tomaslin: probably just 0 * _ here

kevinawoo:
ah i see. Changed!

tomaslin:
LGTM

PR #181, opened 2017-08-16
fix(jenkins): Process brand new builds

jeyrschabu:
- Adjusted algo to account for the first time a build is encountered


igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
```diff
@@ -155,6 +155,7 @@ class JenkinsBuildMonitor implements PollingMonitor {
             log.debug("Jobs on master: ${master} ${jobs.size()}")
 
             for (Project job : jobs) {
+//                if (job.name != "jeyrstest") continue
```
icfantv: I think we can remove this.  ;-)

PR #182, opened 2017-08-16
Pubsub client

jtk54:
This is a start on adding pubsub clients into Igor. Two commits here:

1. Adds relevant configuration, subscription, and monitor classes for Kafka and Google Pubsub. Only Google Pubsub is implemented, Kafka is stubbed in. Handles accounting of received messages in memory.

2. Adds the Redis/Jedis-based client synchronization and uses said implementation in the Google Pubsub message handling.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/pubsub/google/GooglePubsubConfig.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/pubsub/google/GooglePubsubConfig.groovy
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2017 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.igor.config.pubsub.google
+
+import com.netflix.spinnaker.igor.history.EchoService
+import com.netflix.spinnaker.igor.pubsub.PubsubMessageCache
+import com.netflix.spinnaker.igor.pubsub.PubsubSubscribers
+import com.netflix.spinnaker.igor.pubsub.googlePubsub.GooglePubsubSubscriber
+import groovy.transform.CompileStatic
+import groovy.util.logging.Slf4j
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.boot.context.properties.EnableConfigurationProperties
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+
+import javax.validation.Valid
+
+@Configuration
+@Slf4j
+@CompileStatic
+@ConditionalOnProperty("googlePubsub.enabled")
+@EnableConfigurationProperties(GooglePubsubProperties)
+class GooglePubsubConfig {
+
+  @Bean
+  Map<String, GooglePubsubSubscriber> googlePubsubSubscribers(PubsubSubscribers pubsubSubscribers,
+                                                              EchoService echoService,
+                                                              PubsubMessageCache pubsubMessageCache,
+                                                              @Valid GooglePubsubProperties googlePubsubProperties) {
+    log.info("Creating Google Pubsub Subscribers")
+    Map<String, GooglePubsubSubscriber> subscriberMap = googlePubsubProperties?.subscriptions?.collectEntries { GooglePubsubProperties.GooglePubsubSubscription subscription ->
+      log.info "Bootstrapping Google Pubsub Subscriber listening to topic: ${subscription.name} in project: ${subscription.project}"
+      GooglePubsubSubscriber subscriber = GooglePubsubSubscriber
+          .buildSubscriber(subscription.name, subscription.project, subscription.jsonPath, subscription.ackDeadlineSeconds, echoService, pubsubMessageCache)
+      [("projects/${subscription.project}/topics/${subscription.name}".toString()): subscriber]
+    }
+
+    pubsubSubscribers.map.putAll subscriberMap
+    subscriberMap
+  }
+}
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/pubsub/google/GooglePubsubConfig.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/pubsub/google/GooglePubsubConfig.groovy
@@ -18,6 +18,7 @@ package com.netflix.spinnaker.igor.config.pubsub.google
 
 import com.netflix.spinnaker.igor.history.EchoService
 import com.netflix.spinnaker.igor.pubsub.PubsubMessageCache
+import com.netflix.spinnaker.igor.pubsub.PubsubSubscriber
 import com.netflix.spinnaker.igor.pubsub.PubsubSubscribers
 import com.netflix.spinnaker.igor.pubsub.googlePubsub.GooglePubsubSubscriber
 import groovy.transform.CompileStatic
@@ -25,32 +26,41 @@ import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.boot.context.properties.EnableConfigurationProperties
-import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
 
+import javax.annotation.PostConstruct
 import javax.validation.Valid
 
 @Configuration
 @Slf4j
 @CompileStatic
-@ConditionalOnProperty("googlePubsub.enabled")
+@ConditionalOnProperty("pubsub.google.enabled")
 @EnableConfigurationProperties(GooglePubsubProperties)
 class GooglePubsubConfig {
 
-  @Bean
-  Map<String, GooglePubsubSubscriber> googlePubsubSubscribers(PubsubSubscribers pubsubSubscribers,
-                                                              EchoService echoService,
-                                                              PubsubMessageCache pubsubMessageCache,
-                                                              @Valid GooglePubsubProperties googlePubsubProperties) {
+  @Autowired
+  PubsubSubscribers pubsubSubscribers
+
+  @Autowired
+  EchoService echoService
+
+  @Autowired
+  PubsubMessageCache pubsubMessageCache
+
+  @Valid
+  @Autowired
+  GooglePubsubProperties googlePubsubProperties
+
+  @PostConstruct
+  void googlePubsubSubscribers() {
     log.info("Creating Google Pubsub Subscribers")
-    Map<String, GooglePubsubSubscriber> subscriberMap = googlePubsubProperties?.subscriptions?.collectEntries { GooglePubsubProperties.GooglePubsubSubscription subscription ->
+    Map<String, PubsubSubscriber> subscriberMap = googlePubsubProperties.subscriptions?.collectEntries { GooglePubsubProperties.GooglePubsubSubscription subscription ->
       log.info "Bootstrapping Google Pubsub Subscriber listening to topic: ${subscription.name} in project: ${subscription.project}"
       GooglePubsubSubscriber subscriber = GooglePubsubSubscriber
           .buildSubscriber(subscription.name, subscription.project, subscription.jsonPath, subscription.ackDeadlineSeconds, echoService, pubsubMessageCache)
       [("projects/${subscription.project}/topics/${subscription.name}".toString()): subscriber]
     }
 
-    pubsubSubscribers.map.putAll subscriberMap
-    subscriberMap
+    pubsubSubscribers.putAll subscriberMap
   }
 }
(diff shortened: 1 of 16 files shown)


ajordens:
General comments ...

- thoughts on `igor-pubsub` (and/or `igor-pubsub-kafka`, `igor-pubsub-google`) artifacts
- pure java8 (or possibly kotlin) instead of groovy*?

* I realize `igor` is basically 100% groovy right now but it'd be nice to start course correcting that.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/pubsub/google/GooglePubsubConfig.groovy
```diff
@@ -0,0 +1,55 @@
...
+@EnableConfigurationProperties(GooglePubsubProperties)
+class GooglePubsubConfig {
+
+  @Bean
+  Map<String, GooglePubsubSubscriber> googlePubsubSubscribers(PubsubSubscribers pubsubSubscribers,
+                                                              EchoService echoService,
+                                                              PubsubMessageCache pubsubMessageCache,
+                                                              @Valid GooglePubsubProperties googlePubsubProperties) {
+    log.info("Creating Google Pubsub Subscribers")
+    Map<String, GooglePubsubSubscriber> subscriberMap = googlePubsubProperties?.subscriptions?.collectEntries { GooglePubsubProperties.GooglePubsubSubscription subscription ->
```
ajordens: `pubsubSubscribers` shouldn't be null so no need to null dereference it I wouldn't think.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/pubsub/google/GooglePubsubConfig.groovy
```diff
@@ -0,0 +1,55 @@
...
+      log.info "Bootstrapping Google Pubsub Subscriber listening to topic: ${subscription.name} in project: ${subscription.project}"
+      GooglePubsubSubscriber subscriber = GooglePubsubSubscriber
+          .buildSubscriber(subscription.name, subscription.project, subscription.jsonPath, subscription.ackDeadlineSeconds, echoService, pubsubMessageCache)
+      [("projects/${subscription.project}/topics/${subscription.name}".toString()): subscriber]
+    }
+
+    pubsubSubscribers.map.putAll subscriberMap
```
ajordens: Seems like an odd side-effect ... definitely the `pubsubScribers.map.putAll()` looks like an abuse of some internal state.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/pubsub/google/GooglePubsubConfig.groovy
```diff
@@ -0,0 +1,55 @@
...
+
+@Configuration
+@Slf4j
+@CompileStatic
+@ConditionalOnProperty("googlePubsub.enabled")
+@EnableConfigurationProperties(GooglePubsubProperties)
+class GooglePubsubConfig {
+
+  @Bean
+  Map<String, GooglePubsubSubscriber> googlePubsubSubscribers(PubsubSubscribers pubsubSubscribers,
```
ajordens: I'd suggest returning a concrete class vs `Map<String, GooglePubsubSubscriber`>.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/pubsub/google/GooglePubsubProperties.groovy
```diff
@@ -0,0 +1,44 @@
...
+
+package com.netflix.spinnaker.igor.config.pubsub.google
+
+import org.hibernate.validator.constraints.NotEmpty
+import org.springframework.boot.context.properties.ConfigurationProperties
+
+import javax.validation.Valid
+import javax.validation.constraints.NotNull
+
+@ConfigurationProperties(prefix = 'googlePubsub')
```
ajordens: `pubsub.google` a more reasonable config path?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/pubsub/PubsubSubscriber.groovy
```diff
@@ -0,0 +1,26 @@
...
+ */
+
+package com.netflix.spinnaker.igor.pubsub
+
+import com.netflix.spinnaker.igor.model.PubsubType
+import sun.reflect.generics.reflectiveObjects.NotImplementedException
+
+abstract class PubsubSubscriber {
+  static PubsubType pubsubType() {
+    throw new NotImplementedException('Static function "pubsubType()" not implemented')
```
ajordens: `UnsupportedOperationException` IMO vs the sun.* internal

igor-web/src/main/groovy/com/netflix/spinnaker/igor/pubsub/PubsubSubscribers.groovy
```diff
@@ -0,0 +1,27 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.igor.pubsub
+
+import com.netflix.spinnaker.igor.model.PubsubType
+
+class PubsubSubscribers {
+  Map<String, PubsubSubscriber> map = new HashMap<String, PubsubSubscriber>()
```
ajordens: The other side of that internal state leakage commented on above.

Feels like this Bean could be created and provided a `List<PubsubSubscriber>` mebbe.

ajordens:
Thoughts on whether this belongs in `igor` or `echo`?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/pubsub/PubsubSubscribers.groovy
jtk54: The subscribers are recorded by name, emphasized that a bit more in the naming.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/pubsub/google/GooglePubsubProperties.groovy
jtk54: Yep, changed.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/pubsub/PubsubSubscriber.groovy
jtk54: Changed.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/pubsub/google/GooglePubsubConfig.groovy
jtk54: I cleaned up the weird internal state usage and go through proper functions now.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/pubsub/google/GooglePubsubConfig.groovy
jtk54: I actually just want the state change in the `PubsubSubscribers` and not the emitted `Bean`, so I moved this into a constructor for this `Configuration` class.

jtk54:
> thoughts on igor-pubsub (and/or igor-pubsub-kafka, igor-pubsub-google) artifacts
* Could be useful when handling notifications in Echo, but we don't know the full requirements yet. Might feel weird to implement notification logic to be used in Echo in a library in Igor.
> java8 or kotlin
* I can port this to java8, likely not in this PR to keep them small and contained though. I'm not sure how I feel about Kotlin -- seems like it's a drag n drop Groovy replacement and might get replaced with java9 or the next language du jour later on.
> igor or echo
* Other polling clients like Jenkins and Travis are in Igor, and pubsub clients follows that pattern (at least for messages triggering Echo events). This might be less desirable when we introduce outgoing notifications to pubsub, since those would likely live in Echo and would need the same libraries.

lwander:
FWIW kotlin + groovy in the same module is a little tricky as far as dependencies go. When I tried to get this to work I ran into the following:

Both need to transpile to Java first, and in the process they do their dependency resolution. However, since they can't depend on non-java code you can either have only kotlin depending on only groovy, or vice versa depending on your compilation order. Between different modules this is handled for you because you make your dependencies explicit which determines how things are compiled. Within a module it's a little more murky.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/pubsub/PubsubSubscriber.groovy
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+package com.netflix.spinnaker.igor.pubsub
+
+import com.netflix.spinnaker.igor.model.PubsubType
+
+import javax.naming.OperationNotSupportedException
+
+abstract class PubsubSubscriber {
+  static PubsubType pubsubType() {
```
lwander: Is this to distinguish between `PubsubSubscriber` types? If so, why not check the `.class`?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/pubsub/PubsubMessageCache.groovy
```diff
@@ -0,0 +1,74 @@
...
+      resource.psetex(messageKey, retentionDeadlineMillis, identifier)
+    }
+  }
+
+  String makeKey(String messagePayload, PubsubType pubsubType, String subscription) {
+    // NOTE: hashCode() translates a String into a 32-bit integer.
+    // This is relatively small space; however, we are assuming a low
+    // message influx -- tens a minute, which translates to ~10^4 per day.
+    // We persist handled messages for a week maximally, so we assume 7x10^4 messages
+    // accrued per week. However -- 7x10^4 / 2^32 ~ 10^-5 which is sufficiently small
```
lwander: This is pair-wise, have you calculated this with the birthday paradox in mind?

tomaslin:
I'd argue that since the pipeline triggering and cron triggers already live in echo, it's a more natural fit to have this pubsub logic in echo rather than in igor. Conceptually it just replaces the web endpoint for posting events with a pubsub endpoint. Not sure how much value it provides being in igor.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/utils/NodeIdentity.groovy
```diff
@@ -0,0 +1,89 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.igor.utils
+
+import java.lang.management.ManagementFactory
+
+class NodeIdentity {
```
lwander: it reads like if a node dies and comes back up on the same host it will get the same ID - is that OK?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/pubsub/googlePubsub/GooglePubsubSubscriber.groovy
```diff
@@ -0,0 +1,136 @@
...
+                                                PubsubMessageCache pubsubMessageCache) {
+    Subscriber subscriber
+    GooglePubsubMessageReceiver messageReceiver = new GooglePubsubMessageReceiver(
+        ackDeadlineSeconds: ackDeadlineSeconds,
+        echoService: echoService,
+        subscriptionName: name,
+        pubsubMessageCache: pubsubMessageCache
+    )
+
+    if (jsonPath) {
```
lwander: Ah that's what jsonPath is - ignore my above question

jtk54:
@tomaslin We'd have to add the `PollingMonitor` or something similar to allow for polling message queues and add Jedis to Echo. If we're OK with those consequences, this can live in Echo.

> Conceptually it just replaces the web endpoint for posting events with a pubsub endpoint.
Not sure what you mean by this. Seems like we'd avoid using any endpoint if the pubsub clients lived in Echo. To be clear here, the pubsub clients are pulling messages when they are available, not being pushed to. The idea was to separate the polling concerns (this piece in Igor) from the triggering concerns (Echo). Thoughts?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/pubsub/PubsubSubscribers.groovy
```diff
@@ -0,0 +1,35 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.igor.pubsub
+
+import com.netflix.spinnaker.igor.model.PubsubType
+
+class PubsubSubscribers {
+  private Map<String, PubsubSubscriber> subscriberByName = new HashMap<String, PubsubSubscriber>()
```
ajordens: Still on the fence with respects to storing these as a map.

Alternative might be:
```
private Collection<NamedPubsubSubscriber> subscribers

void add(NamedPubsubSubscriber subscriber)
Collection<NamedPubsubSubscriber> getSubscribers(PubsubType pubsubType)
NamedPubsubSubscriber getByName(String name)
```

I generally find it easier to filter and manipulate lists rather than maps where you might need to look at both the key and value.

My feelings aren't terribly strong so I won't be upset if you keep this `Map`-based.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/pubsub/googlePubsub/GooglePubsubMonitor.groovy
```diff
@@ -0,0 +1,99 @@
...
+        { subscription -> closeConnection(subscription) }
+    )
+  }
+
+  @Override
+  void onApplicationEvent(ContextRefreshedEvent event) {
+    // TODO(jacobkiefer): Register Igor as enabled on startup.
+    log.info('Starting async connections for Google Pubsub subscribers')
+    pubsubSubscribers.filteredMap(PubsubType.GOOGLE).keySet().parallelStream().forEach(
+        { subscription -> openConnection(subscription) }
```
ajordens: might need some exception handling (here and `closeAsyncConnections()`)?

jtk54:
Update on this -- porting to java8 and moving it to its own submodule in Echo. Closing this one. I'll incorporate the comments in the new implementation.

PR #187, opened 2017-09-12
chore(jenkins): Make logs less chatty

jeyrschabu:
- Only log on actions, removed some logs causing noise


ajordens:
Seems alright, haven't had to look at igor logs in awhile so if we don't think these are valuable when debugging lets get this merged!

PR #188, opened 2017-09-19
feat(authz): Add ability to specify Google service account credential…

ttomsu:
… or raw OAuth bearer token to Jenkins master request.

I'm not 100% sure this will actually work as expected. Most Jenkins masters with username/password protection (including the one I tested with) require the `Basic` header, so any Oauth-guarded proxy will also require the ability to discern multiple values in the `Authorization header`. 

According to https://stackoverflow.com/a/38515091/5569046, this should work by RFC spec, but may not actually work in the wild. (cc @katrielt )


duftler:
LGTM.
Would like @tomaslin to weigh in as well.

skim1420:
@tomaslin ping thx

ttomsu:
@tomaslin Anything to add here? If not I'll merge.

robzienert:
@tomaslin ping - last chance

PR #189, opened 2017-10-18
feat(artifact) This will enable the decoration of artifacts by default

brujoand:
As a step towards improved artifact handling it would be useful to now enable the artifact decoration by default so that Igor also will be able to align with the work being done on this in Orca and Rosco.


brujoand:
@tomaslin the idea here is to enable the decoration by default, so that we can start adapting orca to handle the decorated artifacts. Does this change make sense to you?

jeyrschabu:
Hey @gardleopard could you expand on the impact of this change?

gardleopard:
@jeyrschabu this will allow us to assume that all artifacts coming from CI are decorated with version and type. This will pave the way for deprecating this https://github.com/spinnaker/orca/blob/fc87a8e99ec59a4e129646026987aa5adbde2d31/orca-core/src/main/java/com/netflix/spinnaker/orca/pipeline/util/PackageInfo.java#L33 

When we have this we can make a feature toggle flag for skipping the packageInfo part in orca and simplify the logic for triggering a bake stage.

robzienert:
@gardleopard Is this still needed? Waiting for someone to review / merge?

lwander:
@ezimanyi fyi for existing artifact decoration work

gardleopard:
For us this is already enabled, it depends on the adoption of decorated artifacts if this should be enabled by default or not imho.

robzienert:
I'll defer to @lwander or @ezimanyi on merge for this PR.

ezimanyi:
I think that it probably makes sense to encourage users to adopt Jenkins artifacts as added in #401 instead of enabling this by default.  Any objections to closing this @gardleopard.

PR #191, opened 2017-10-20
feat(travis/commit_info): Add commit timestamp to build info

jervi:
Adds a `timestamp` property to scm information from Travis.
```json
    "scm": [
        {
            "branch": "master",
            "committer": "Jørgen Jervidalo",
            "compareUrl": "https://github.com/org/repo/compare/5f03eddcdd7c...4b87d9976902",
            "timestamp": "2017-10-20T14:09:13Z",
            "message": "Commit message",
            "name": "master",
            "sha1": "4b87d997690295d97befd9d285e990603a2e8147"
        }
    ]
```


jervi:
Currently we are not using the timestamp in Igor, so it's just serialized back to a String. Maybe I should just pass it through as a String and remove all the converters?
The upside of using converters is that we're sure the format of the timestamp passed to Orca will always be in ISO-8601, and it might be useful when implementing support for other CI systems etc. The Jenkins integration is not currently populating the field.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/TravisConfig.groovy
```diff
@@ -79,6 +84,10 @@ class TravisConfig {
             .setClient(new OkClient(client))
             .setLog(fooLog)
             .setLogLevel(RestAdapter.LogLevel.FULL)
+            .setConverter(new GsonConverter(new GsonBuilder()
+                .registerTypeAdapter(Instant.class, new InstantTypeConverter())
+                .registerTypeAdapter(Instant.class, new InstantTypeConverter())
```
amanya: Needs to be run twice?

amanya:
I put a comment but LGTM

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/TravisConfig.groovy
jervi: Not at all! Will fix

jervi:
I've added another commit here that is meant to clean things up a bit. The travis integration was, as the only integration in Igor, using Gson for serialization (through Retrofit), but it is using Jackson for deserialization (when requests are made from Orca). I've removed Gson and switched to Jackson in both ends.

igor-web/igor-web.gradle
```diff
@@ -19,8 +19,8 @@ dependencies {
     compile spinnaker.dependency("kork")
     compile spinnaker.dependency("korkStackdriver")
     compile spinnaker.dependency("korkWeb")
+    compile spinnaker.dependency("jacksonJsr310")
     compile spinnaker.dependency("jedis")
-    compile spinnaker.dependency("guava")
```
jervi: Guava is never used in Igor

jeyrschabu:
@jervi can you update this PR? It looks good otherwise

jervi:
Done

PR #196, opened 2017-11-18
fix(jenkins): adding retry for gitDetails

emjburns:
Saw an error where the scm information was not pulled from jenkins during a pipeline. A re-run of the pipeline fixed this. I think adding a retry for this call to handle any jenkins blips might solve the problem. 

Thoughts? @jeyrschabu @robzienert

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/service/JenkinsService.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/service/JenkinsService.groovy
@@ -22,20 +22,18 @@ import com.netflix.spinnaker.igor.build.BuildController
 import com.netflix.spinnaker.igor.build.model.GenericBuild
 import com.netflix.spinnaker.igor.build.model.GenericGitRevision
 import com.netflix.spinnaker.igor.jenkins.client.JenkinsClient
-import com.netflix.spinnaker.igor.jenkins.client.model.Build
-import com.netflix.spinnaker.igor.jenkins.client.model.BuildDependencies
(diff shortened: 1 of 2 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/service/JenkinsService.groovy
```diff
@@ -126,10 +129,12 @@ class JenkinsService implements BuildService{
...
-        new SimpleHystrixCommand<ScmDetails>(
-            groupKey, buildCommandKey("getGitDetails"), {
-            return jenkinsClient.getGitDetails(encode(jobName), buildNumber)
-        }).execute()
+        retrySupport.retry({
+            new SimpleHystrixCommand<ScmDetails>(
+                groupKey, buildCommandKey("getGitDetails"), {
+                return jenkinsClient.getGitDetails(encode(jobName), buildNumber)
+            }).execute()
+        }, maxRetries, retryBackoff, exponentialBackoff)
```
ajordens: `retryBackoff` is in milliseconds, I would probably not do exponential either and just favor more calls in shorter-timeframe vs less calls over longer timeframe (maybe 10 retries, 1-2000ms backoff or something?)

Not a huge amount of value defining the `maxRetries`/`retryBackoff`/`exponentialBackoff` at the top of this class ... they're only used in one spot and not externally configurable.

emjburns: Makes sense! Fixed.

PR #198, opened 2017-11-22
feat(jenkins): Adding a configurable lookBack window, off by default

jeyrschabu:
- When turned on, builds older than `lookBackWindowMins` will not be processed
- offsetting that with the polling interval


jeyrschabu:
This will prevent a triggering old builds in the case, the backing store (redis), is flushed or is terminated.

To turn this on:
```
spinnaker:
  build:
    lookBackWindowMins: 30
    processBuildsOlderThanLookBackWindow: false

```

jeyrschabu:

jeyrschabu:
Merging this for now. will look at Docker next and generalize

PR #201, opened 2017-12-01
fix(buildevents): lookback window checks end time, improved logs

emjburns:
With the new lookback window, we were checking if the start time was before the window, and if so ignoring it. This caused builds that were longer than the lookback window to not to fire an event.

This changes the calculation to check whether the build end time is outside the lookback window.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
@@ -30,6 +30,7 @@ import com.netflix.spinnaker.igor.polling.PollingMonitor
 import com.netflix.spinnaker.igor.service.BuildMasters
 import groovy.time.TimeCategory
 import groovy.util.logging.Slf4j
+import org.apache.commons.lang3.time.DateUtils
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.beans.factory.annotation.Value
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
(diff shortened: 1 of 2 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
```diff
@@ -190,14 +191,22 @@ class JenkinsBuildMonitor implements PollingMonitor {
...
-                                Date buildDate = new Date(it.timestamp as Long)
-                                log.info("filtering builds older than {}", lookBackDate)
-                                return buildDate.after(lookBackDate)
+                            def offsetSeconds = pollInterval.seconds
+                            def lookBackWindowMins = igorConfigurationProperties.spinnaker.build.lookBackWindowMins.minutes
+                            Date lookBackDate = (offsetSeconds + lookBackWindowMins).ago
+
+                            def tooOldBuilds = completedBuilds.findAll {
+                                Date buildEndDate = new Date(it.timestamp as Long)
+                                DateUtils.addMilliseconds(buildEndDate,Math.toIntExact(it.duration))
```
ajordens: I've never used `DateUtils.addMilliseconds()` but the javadoc indicates it returns a `Date`, we don't appaer to be using it?

@emjburns ^^

emjburns: ya good feedback. I"ll change to that. it.duration is the number of ms the build took)

igor-web/src/test/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitorSpec.groovy
```diff
@@ -174,26 +174,27 @@ class JenkinsBuildMonitorSpec extends Specification {
...
-        long stamp1 = System.currentTimeMillis() - (10 * 60 * 1000) // 10 minutes ago
-        long stamp2 = System.currentTimeMillis() - (5 * 60 * 1000)  // 5  minutes ago
-        long stamp3 = System.currentTimeMillis()
+        long now = System.currentTimeMillis()
+        long nowMinus30min = now - (30 * 60 * 1000) // 30 minutes ago
+        long nowMinus10min = now - (10 * 60 * 1000) // 10 minutes ago
+        long nowMinus5min = now - (5 * 60 * 1000)  // 5  minutes ago
+        long durationOf5min = 300000
 
         and: 'a 6 minutes total lookBack window'
```
ajordens: this title isn't quite right anymore ... it's now `65` minutes right?

emjburns: Should be a 6 minute lookback window. The poll interval is 60 seconds, which is now correctly being translated for the lookback window calculation. In this test, the lookBackWindow is set to 5 minutes. Together, that gives 6 minutes.

PR #203, opened 2017-12-06
feat(web): RedisClientDelegate for Dynomite support

robzienert:
* Updates gradlew version to 3.5
* Update to latest spinnaker-dependencies
* Refactored services that referenced `Jedis` to use `RedisClientDelegate` instead
* Classes touched (Spring Config aside) converted to Java
* Refactored Spec classes touched to remove Spring dependency so things run faster
* Added Dynomite configuration class

cc @spinnaker/reviewers


igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/DynomiteConfigurationProperties.groovy
```diff
@@ -0,0 +1,68 @@
...
+        String hostname
+
+        String ipAddress
+
+        int port = Host.DEFAULT_PORT
+
+        Host.Status status = Host.Status.Up
+
+        @NotNull
+        String rack = 'localrack'
```
jrsquared: why not use above defined `localRack` and `localDataCenter`?

igor-web/src/test/groovy/com/netflix/spinnaker/igor/build/BuildCacheSpec.groovy
```diff
@@ -79,7 +59,7 @@ class BuildCacheSpec extends Specification {
         cache.getLastBuild(master, 'job1', true) == 80
     }
 
-    void 'running and completed builds are handled seperatly'() {
+    void 'running and completed builds are handled separately'() {
```
jrsquared: best change

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/DynomiteConfigurationProperties.groovy
robzienert: They're different. `localRack` and `localDataCenter` define the rack & dc of the `igor` instance, whereas `DynoHost` are static pointers to dynomite hosts, which may not be in the same dc or rack.

These happen to have the same default values because we've never needed to statically define anything except during local development, but the values are still required by the dyno client for connection pooling, etc.

PR #204, opened 2017-12-06
fix(discovery): Fixing cyclic dependency issue on startup when discovery client avail

robzienert:
When running igor with Discovery & Dynomite enabled at the same time, we run into a pretty awesome cycle:

```
   buildCache defined in URL [jar:file:/BuildCache.class]
┌─────┐
|  dynomiteClientDelegate defined in class path resource [DynomiteConfig.class]
↑     ↓
|  dynoJedisClient defined in class path resource [DynomiteConfig.class]
↑     ↓
|  discoveryClient defined in class path resource [EurekaComponents.class]
↑     ↓
|  optionalArgs defined in class path resource [EurekaComponents.class]
↑     ↓
|  healthCheckHandler defined in class path resource [EurekaComponents.class]
↑     ↓
|  pollingMonitorHealth (field private PollingMonitorHealth.pollers)
↑     ↓
|  dockerMonitor (field private DockerMonitor.cache)
↑     ↓
|  dockerRegistryCache defined in URL [jar:file://DockerRegistryCache.class]
└─────┘
```

Despite `DockerMonitor` using a provided DiscoveryClient, the class listens on all `ContextRefreshedEvent`, which happens quite a lot in startup. Rather than scheduling a discovery poller on every refresh event, altered the monitors to use `RemoteStatusChangedEvent` instead, which should offer the same behavior, but also let wiring happen without cycles.

cc @cfieber 

😍 Cameron for showing me `-Ddebug`

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/NoDiscoveryApplicationStatusPublisher.java
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/NoDiscoveryApplicationStatusPublisher.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2017 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


cfieber:
LGTM - if it matters we could also start/stop pollers on UP vs OUT_OF_SERVICE - not sure if that gives us a better deploy story.

But that can be a task for another day..

robzienert:
That's effectively what this does, it'll make igor behave like orca does with respect to discovery status. The magical goodness that handles that is mostly in kork.

PR #197, opened 2017-11-20
feat(gitlab): Add Gitlab SCM integration (spinnaker/spinnaker#2047)

mdirkse:
Enables Igor scm functionality for GitLab (as per spinnaker/spinnaker#2047)

PR #207, opened 2017-12-08
WIP: feat(gitlab CI integration) (spinnaker/spinnaker#2047)

wheleph:
**This is a work-in-progress PR. Do not merge.**

This is the part of https://github.com/spinnaker/spinnaker/issues/2047 and the follow-up for https://github.com/spinnaker/igor/pull/197 to not only the SCM but also the CI aspect of Gitlab.

This is a work-in-progress implementation of support for Gitlab CI, namely:
- triggering of Spinnaker pipelines after completions of Gitlab CI pipelines
- pulling artifacts from Gitlab CI pipelines for bake stages
- Triggering of Gitlab CI pipelines from Spinnaker steps

I've been working on the Gitlab CI integration: https://github.com/spinnaker/igor/pull/207. This is still work-in-progress but I'd like to have some input because the PR is growing bigger and bigger. Mainly because I felt like I needed to do some refactoring:
- Changing type of build number from int to long since for gitlab.com they apparently use a global id counter for all pipelines and it's not unrealistic that there'll be an integer overflow in the near future. So it's better to be on the safe side
- Extracting of common parts polling monitors into CommonPollingMonitor
- Defining some common methods into BuildService interface

So I have the following questions to you guys:
- What's the general feeling about the PR? Should we keep it as a big one or is it better to split it into several parts?
- Since the change touches the existing integrations (Travis, Jenkins, Docker etc) what would be the best way to test it?
- Also changes in other services will be needed (orca, gate, deck) but I believe that they should be in separate PRs.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildCache.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildCache.groovy
@@ -54,14 +54,14 @@ class BuildCache {
         results
     }
 
-    int getLastBuild(String master, String job, boolean running) {
+    long getLastBuild(String master, String job, boolean running) {
         Jedis resource = jedisPool.resource
         def key = makeKey(master, job, running)
         if (!resource.exists(key)) {
             jedisPool.returnResource(resource)
             return -1
         }
-        int buildNumber = resource.get(key) as Integer
+        int buildNumber = resource.get(key) as Long
 
         jedisPool.returnResource(resource)
         buildNumber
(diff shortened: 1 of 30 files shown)


jeyrschabu:
The direction is reasonable. This is a rather large PR and should probably be split into smaller digestible ones. 
One way to reduce the blast radius would be to have a PR that addresses adding GitLab CI first, then port the abstractions to generalize the current implementation i.e jenkins polling ect..

wheleph:
@jeyrschabu agree about the size of the PR (it's even not complete yet!). But I would rather create a PR for the refactoring first since this way we'll be able to keep the codebase of Igor consistent over at every point of time.

jeyrschabu:
@wheleph that's probably fine as long as the PR's are small.

PR #209, opened 2017-12-13
refactor(core): Reduce code duplication in polling monitors

wheleph:
This PR fixes https://github.com/spinnaker/spinnaker/issues/2166.

The basic idea is to reduce the code duplication by using [template method design pattern](https://en.wikipedia.org/wiki/Template_method_pattern). All the common code is pushed to `CommonPollingMonitor`.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerMonitor.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerMonitor.groovy
@@ -16,37 +16,22 @@
 
 package com.netflix.spinnaker.igor.docker
 
-import com.netflix.appinfo.InstanceInfo
-import com.netflix.discovery.DiscoveryClient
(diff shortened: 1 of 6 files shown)


jeyrschabu:
Looks good. One comment I have is that we prefer Java for new classes.

wheleph:
I see. I guess it means that I need to reimplement the whole hierarchy of build monitors in Java since it will be a bit confusing to have the parent class in Java and the rest in Groovy. It shouldn't be very hard and I'm going to work on this.

But I'm wondering why are you going away from Groovy?

Also I assume it doesn't mean that we need to re-implement the test in Java, right?

jeyrschabu:
@wheleph it's a new pattern we are adopting (favor java over groovy for new classes). 

That's a good point in this situation with the child classes being of a different type. No need to convert the class

wheleph:
If you're cool with leaving the code in Groovy then it's ok. Do you have any idea when the code will be merged into master so that I can continue with the next PR?

PR #210, opened 2017-12-14
Revert "refactor(cache): Replace usage of redis keys command with scan (#208)"

robzienert:
`scan` is not correctly implemented in the dyno client. Projected fix date is Q1 2018 - so will be reverting. This doesn't affect Redis clients.


sushantk:
@robzienert - is the scan fix ready to go again if implemented by dyno client? We can certainly use it to solve a nagging latency issue due to keys command slowing down redis significantly.

robzienert:
@sushantk I haven't gone back to fix this, no. Thanks for the reminder, this had slipped my mind and I can pretty easily re-add this functionality. Will try to get to it this week.

PR #219, opened 2018-01-23
fix(jenkins): Continue to process builds from a master if a single build fetch fails

christopherthielen:
Previously, if one job failed to process during `changedBuilds`, no other jobs from that build master would be processed. 

This wraps the individual job with a try/catch and logs the retrofit URL that failed, if any.


christopherthielen:
Use `?w=1` to view diff/ignore whitespace https://github.com/spinnaker/igor/pull/219/files?w=1

PR #222, opened 2018-02-09
feat(jenkins): Adds csrf crumb config and handling.

jtk54:


igor-web/src/main/groovy/com/netflix/spinnaker/igor/model/Crumb.groovy
```diff
@@ -0,0 +1,16 @@
+package com.netflix.spinnaker.igor.model
```
lwander: Missing license header

jtk54:
https://github.com/spinnaker/spinnaker/issues/2067.

jtk54:
@tomaslin merging if no complaints.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/JenkinsProperties.groovy
```diff
@@ -45,6 +45,8 @@ class JenkinsProperties {
         @NotEmpty
         String password
 
+        Boolean preventCSRF = false
```
tomaslin: can we just change this to `csrf` and have it be true when it is enabled and false when not? `Cross-Site Request Forgery = true / false` is more user friendly.

tomaslin:
LGTM

PR #223, opened 2018-02-22
feat(travis): Include artifacts in build event for completed builds

jervi:
This is needed for filtering build events based on artifacts. Depends on an upcoming change in echo.
Kind of a duplicate of #132, but this change will only query Travis for artifacts for completed builds, and only once per event.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/client/model/v3/V3Log.java
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/client/model/v3/V3Log.java
@@ -1,0 +1,82 @@
+/*
+ * Copyright 2018 Schibsted ASA.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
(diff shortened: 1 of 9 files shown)


jervi:
Wait a minute. Sorting out some edge cases in the Travis' API

igor-web/src/main/groovy/com/netflix/spinnaker/igor/travis/client/model/v3/V3Log.java
```diff
@@ -0,0 +1,82 @@
...
+
+  public Integer getId() {
+    return id;
+  }
+
+  public void setId(Integer id) {
+    this.id = id;
+  }
+
+  public String getContent() {
```
jervi: Turns out that the `content` field isn't yet populated right after a build completes. Instead, the log is split over several `logParts` fields. If you do the same API call a few minutes later, these `logParts` will have been concatenated and put into the `content` field.
With this commit, we support both ways.

jervi:
Echo PR is also ready: https://github.com/spinnaker/echo/pull/240

jervi:
@spinnaker/reviewers

PR #224, opened 2018-02-23
refactor(jenkins): mostly cosmetic changes

dreynaud:
* rename changedBuilds to processProjectsOfMaster
* break processProjectsOfMaster into multiple functions to make it
  more readable and decrease indentation levels
* use early exits when possible
* make things private when possible
* don't explicitly build the list of jobs too old to include only to
  exclude it from the main list (side effect: no longer debug logging
  them)
* things should otherwise work in the same way

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
@@ -78,7 +78,7 @@ class JenkinsBuildMonitor extends CommonPollingMonitor {
     void poll() {
         log.info "- Polling cycle started - ${new Date()}"
         buildMasters.filteredMap(BuildServiceProvider.JENKINS).keySet().parallelStream().forEach(
-            { master -> changedBuilds(master) }
+            { master -> processProjectsOfMaster(master) }
         )
         log.info "- Polling cycle done - ${new Date()}"
     }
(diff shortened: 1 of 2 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
```diff
@@ -78,7 +78,7 @@ class JenkinsBuildMonitor extends CommonPollingMonitor {
     void poll() {
         log.info "- Polling cycle started - ${new Date()}"
         buildMasters.filteredMap(BuildServiceProvider.JENKINS).keySet().parallelStream().forEach(
-            { master -> changedBuilds(master) }
+            { master -> processProjectsOfMaster(master) }
```
jeyrschabu: Did we rename in the Monitor interface? not seeing it in this changeset

jeyrschabu:
I like the idea 👍

igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
robzienert: The monitor interface doesn't really define much.

dreynaud: What Rob said, this is not part of the interface

robzienert:
Ping. Would like to see this updated!

dreynaud:
Fixed the conflicts on this branch: https://github.com/spinnaker/igor/pull/238

PR #227, opened 2018-02-27
feat(docker): Redis key migration

robzienert:
Updates redis key format from v1 to v2 for docker registry cache, which removes the docker repository url from the key.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerRegistryCacheV2KeysMigration.java
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerRegistryCacheV2KeysMigration.java
@@ -1,0 +1,180 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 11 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerRegistryCacheV2KeysMigration.java
```diff
@@ -0,0 +1,180 @@
...
+            result = jedis.scan(cursor, params);
+            keys.addAll(result.getResult().stream().filter(this::isOldKey).collect(Collectors.toList()));
+        } while (!result.getStringCursor().equals("0"));
+
+        return keys;
+    }
+
+    private boolean isOldKey(String key) {
+        // Target v1 keys. They include repository URLs, which can include ports, so eq/gt 6 parts.
+        return key.split(":").length >= 6;
```
dreynaud: are we certain this can't fire on new keys? For instance, we could check that the 2nd token is not "v2" as a safety

igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerRegistryCacheV2KeysMigration.java
```diff
@@ -0,0 +1,180 @@
...
+    }
+
+    private String convertToNewFormat(String oldKey) {
+        String[] tagParts = oldKey.split("/");
+        String[] parts = tagParts[0].split(":");
+
+        String account = parts[0];
+        String registry = parts[1];
```
dreynaud: would it make sense to add a length check here before accessing `parts[0]` and `parts[1]`?

robzienert: Yep, 100% plan to.

PR #228, opened 2018-02-28
POC: vanilla gradle

cfieber:
just a proof of concept of stripping out some of the extensions we use on gradle and moving to a vanilla buildscript.

This could subsequently be extended by an external .gradle file included via gradle CLI (-I somefile.gradle) to add customized publishing.

This currently has no external buildscript dependencies (no non-core gradle plugins) but I'd expect we would still bring in the ospackage plugin to produce debs/rpms.

I also think once we are happy with the experience there is value in extracting the conventions back out into a simpler spinnaker gradle plugin instead of copy and pasting them in every project.

This build has no opinion on versioning (can be provided via -Pversion=whatever) or publishing (it does have the maven plugin so you can do an `install` to publish to maven local)

Also fwiw it doesn't bring in the spring-boot plugin (that conflates both the bootRun bit which is unnecessary given the application run task, and some dependency management stuff which is maybe a bit too much magic for what it provides)

### DIFF
--- a/igor-web/igor-web.gradle
+++ b/igor-web/igor-web.gradle
@@ -1,51 +1,109 @@
+
 ext {
   springConfigLocation = System.getProperty('spring.config.location', "${System.getProperty('user.home')}/.spinnaker/")
-  repackage = System.getProperty('springBoot.repackage', "false")
+  versions = [
+    cglib: '3.2.0',
+    kork: '1.124.0',
+    eureka: '1.8.6',
+    groovy: '2.4.13',
+    hamcrest: '1.3',
+    jackson: '2.9.2',
+    junit: '4.12',
+    logstash: '4.11',
+    objenesis: '2.5.1',
+    okHttp: '2.7.0',
+    retrofit: '1.9.0',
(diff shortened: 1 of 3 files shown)


igor-web/igor-web.gradle
```diff
@@ -1,51 +1,108 @@
...
+    logstash: '4.11',
+    objenesis: '2.5.1',
+    okHttp: '2.7.0',
+    retrofit: '1.9.0',
+    rxJava: '1.0.16',
+    snakeyaml: '1.15',
+    spock: '1.1-groovy-2.4',
+    spring: '4.3.14.RELEASE',
+    springBoot: '1.5.10.RELEASE',
+  ]
```
robfletcher: Wouldn't you want these to be up a level so you could sync across modules?

cfieber: Yeah definitely - there's only one module in Igor   so I was lazy as I was unwinding the spinnaker.dependencies notation

robzienert:
Oh man. This is delicious.

PR #232, opened 2018-03-15
feat(jenkins): Adding interface to supply custom OkHttpClient

robzienert:
This is to support providing our own `SSLContextFactory`. Since we're going to be migrating Jenkins one-by-one, I wanted to provide an interface that we could provide different OkHttpClients per-host.

This does not change existing OSS functionality.


cfieber:
LGTM

There is a custom okhttp client now in internal gate and orca if and when
it comes time to leverage this.

PR #234, opened 2018-03-16
chore(*): Add logging for key migration progress

robzienert:
Existing logging doesn't really give much (any) insight into how we're progressing through the key migration.


emjburns:
LGTM

PR #235, opened 2018-03-16
refactor(docker): Incrementally migrate keys on each scan result

robzienert:
It occurred to me while running the migration that if network traffic interrupts the scan process, we've done a bunch of work without any actual output. This refactor is more resilient for Spinnaker deployments that are huge.

Changes:

1. Scan operations now retry on errors (wow!)
1. Each scan operation's results are migrated immediately, rather than being aggregated and processed in a second phase.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerRegistryCacheV2KeysMigration.java
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerRegistryCacheV2KeysMigration.java
@@ -17,6 +17,7 @@ package com.netflix.spinnaker.igor.docker;
 
 import com.google.common.collect.Iterables;
 import com.netflix.dyno.connectionpool.CursorBasedResult;
+import com.netflix.dyno.connectionpool.exception.DynoException;
 import com.netflix.dyno.jedis.DynoJedisClient;
 import com.netflix.spinnaker.igor.IgorConfigurationProperties;
 import com.netflix.spinnaker.kork.dynomite.DynomiteClientDelegate;
@@ -30,6 +31,7 @@ import redis.clients.jedis.Jedis;
 import redis.clients.jedis.MultiKeyCommands;
 import redis.clients.jedis.ScanParams;
 import redis.clients.jedis.ScanResult;
+import redis.clients.jedis.exceptions.JedisException;
 import rx.Scheduler;
 import rx.schedulers.Schedulers;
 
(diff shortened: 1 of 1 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerRegistryCacheV2KeysMigration.java
```diff
@@ -131,48 +125,75 @@ private void migrateBatch(List<DockerRegistryV1Key> oldKeys) {
...
-        CursorBasedResult<String> result;
+        CursorBasedResult<String> result = null;
         do {
-            result = dyno.dyno_scan(pattern);
-            keys.addAll(oldKeysCallback.apply(result.getResult()));
-        } while (!result.isComplete());
+            try {
+                result = dyno.dyno_scan(pattern);
+                numMigrated += oldKeysCallback.apply(result.getResult());
+            } catch (DynoException e) {
```
emjburns: Should we keep retrying forever? Are there some exceptions that mean we don't want to retry?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerRegistryCacheV2KeysMigration.java
```diff
@@ -131,48 +125,75 @@ private void migrateBatch(List<DockerRegistryV1Key> oldKeys) {
...
+        ScanResult<String> result = null;
         do {
-            result = jedis.scan(cursor, params);
-            keys.addAll(oldKeysCallback.apply(result.getResult()));
-        } while (!result.getStringCursor().equals("0"));
+            try {
+                result = jedis.scan(cursor, params);
+                numMigrated += oldKeysCallback.apply(result.getResult());
+            } catch (JedisException e) {
+                log.error("Failed migrating v1 key batch, retrying", e);
```
emjburns: Same comment, do we want to retry forever ever second?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerRegistryCacheV2KeysMigration.java
robzienert: Good point. I've added a failure counter to abort if there's an excessive amount of consecutive failures.

emjburns:
LTGM

PR #238, opened 2018-03-19
refactor(jenkins): mostly cosmetic changes

dreynaud:
break processProjectsOfMaster into multiple functions to make it
more readable and decrease indentation levels
use early exits when possible
make things private when possible
don't explicitly build the list of jobs too old to include only to
exclude it from the main list (side effect: no longer debug logging
them)
things should otherwise work in the same way


dreynaud:
@robzienert this is just https://github.com/spinnaker/igor/pull/224 reapplied on top of your changes

PR #239, opened 2018-03-21
fix(travis): Fix a regression that cause getBuilds() to be very slow

jervi:
The code was always _supposed_ to fetch logs in this code path. However, due to a bug in the JSON to Groovy deserialization, the field containing job id's was always null. Due to groovy's nature, this just caused Igor to not fetch any logs, instead of failing.
I fixed the issue (in #223), but then realized that the log fetching made the code way slower in places where we didn't really need artifacts. So I've disabled logs/artifacts fetching from everywhere but the TravisBuildMonitor.


jervi:
@spinnaker/reviewers PTAL

PR #242, opened 2018-03-23
fix(polling): Fix igor startup with no active ci accounts

ezimanyi:
Currently igor fails on startup when there are no active CI accounts,
due to a spring dependency error.


ezimanyi:
@robzienert : Mind taking a look?  As of https://github.com/spinnaker/igor/pull/226, igor is crashing on start-up if there are no CI accounts configured as it can't find any beans of type `CommonPollingMonitor` (and the behavior is not to fall back to an empty list).  This explicitly makes it fall back on an empty list.

robzienert:
Neat, thanks! Sorry about that.

PR #247, opened 2018-03-28
feat(jenkins): extension point for http request interceptor

cfieber:
Allows customization of the Jenkins request interceptor for http calls.
Makes username/password non manditory on jenkins host configuration.

Example use case would be x509 authentication in Jenkins where the identity
is extracted from the client certificate instead of an HTTP request header.


ajordens:
LGTM

PR #248, opened 2018-04-03
fix(polling): Regression causes send event logic to be inverted

jervi:
#241 lead to Travis and Gitlab no longer sending events when they should (and sending events when they should not). Also, the send event flag was ignored for Jenkins; this commit also fixes that.

FYI @robzienert


igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
```diff
@@ -93,7 +93,7 @@ class JenkinsBuildMonitor extends CommonPollingMonitor<JobDelta, JobPollingDelta
         long startTime = System.currentTimeMillis()
         log.info "Polling cycle started: ${new Date()}"
         buildMasters.filteredMap(BuildServiceProvider.JENKINS).keySet().parallelStream().forEach(
-            { master -> pollSingle(new PollContext(master)) }
+            { master -> pollSingle(new PollContext(master, !sendEvents)) }
```
jervi: I'm just guessing that the intention is to use the flag for Jenkins as well?

robzienert:
Derp.

PR #249, opened 2018-04-03
fix(health): non fatal healthindicator

cfieber:
Updates the PollingMonitor HealthIndicator to no longer
mark the service as DOWN if a PollingMonitor has not
updated recently enough.

Typical deployment footprint for igor is a single instance
due to polling coordination state, so taking that instance
DOWN on any one monitor failure is bad for all clients.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/health/PollingMonitorHealth.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/health/PollingMonitorHealth.groovy
@@ -27,6 +27,8 @@ import org.springframework.boot.actuate.health.Status
 import org.springframework.context.ApplicationContext
 import org.springframework.stereotype.Component
 
+import java.util.concurrent.atomic.AtomicBoolean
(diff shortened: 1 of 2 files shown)


ajordens:
LGTM

igor-web/src/main/groovy/com/netflix/spinnaker/igor/health/PollingMonitorHealth.groovy
```diff
@@ -46,8 +50,9 @@ public class PollingMonitorHealth implements HealthIndicator {
                 if (poller.lastPoll == null) {
                     healths << Health.unknown().withDetail("${poller.name}.status", 'not polling yet').build()
                 } else {
-                    // Check if twice the polling interval has elapsed.
```
ajordens: this comment almost matched what the code was doing

cfieber: hah

PR #255, opened 2018-04-05
feat(docker): Fallback value for item circuit breaker in Docker poller

jervi:
Make it possible to set an `itemUpperThreshold` property for the Docker Registry poller in Igor, that will function as a fallback value if the Docker registries retrieved from Clouddriver does not have this value set.

@robzienert PTAL

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerMonitor.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerMonitor.groovy
@@ -20,6 +20,7 @@ import com.netflix.discovery.DiscoveryClient
 import com.netflix.spectator.api.Registry
 import com.netflix.spinnaker.igor.IgorConfigurationProperties
 import com.netflix.spinnaker.igor.build.model.GenericArtifact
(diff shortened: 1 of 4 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerMonitor.groovy
```diff
@@ -177,7 +181,11 @@ class DockerMonitor extends CommonPollingMonitor<ImageDelta, DockerPollingDelta>
 
     @Override
     protected Integer getPartitionUpperThreshold(String partition) {
-        return dockerRegistryAccounts.accounts.find { it.name == partition }?.itemUpperThreshold
+        def upperThreshold = dockerRegistryAccounts.accounts.find { it.name == partition }?.itemUpperThreshold
```
jervi: As far as I can tell, this property will never exist?
https://github.com/spinnaker/clouddriver/blob/master/clouddriver-docker/src/main/groovy/com/netflix/spinnaker/clouddriver/docker/registry/security/DockerRegistryNamedAccountCredentials.groovy

PR #257, opened 2018-04-05
fix(docker): Fix igor startup with dockerRegistry disabled

ezimanyi:


ezimanyi:
#255 broke startup of igor if `dockerRegistry` is disabled; I think `@ConditionalOnClass` should instead be `@ConditionalOnBean`

jervi:
Oh, sorry abut this! My bad 😬

PR #265, opened 2018-04-24
fix(jenkins) Don't retry conversion errors in getGitDetails

willgorman:
The addition of retries to JenkinsService (https://github.com/spinnaker/igor/pull/196) has exposed a latent issue for us in JenkinsService.

`JenkinsService#getGitDetails` only expects the response to have a single `<action>` element but we've found that it's possible for it to have multiple, like:

```
<workflowRun _class="org.jenkinsci.plugins.workflow.job.WorkflowRun">
  <action _class="hudson.plugins.git.util.BuildData">
    <lastBuiltRevision>
      <branch>
        <SHA1>dc826d8baa540586762f3106be67d9468ab589e6</SHA1>
        <name>master</name>
      </branch>
    </lastBuiltRevision>
  </action>
  <action _class="hudson.plugins.git.util.BuildData">
    <lastBuiltRevision>
      <branch>
        <SHA1>c5fef31e36f30cda898b0464b71e80230aa86cdf</SHA1>
        <name>eng</name>
      </branch>
    </lastBuiltRevision>
  </action>
</workflowRun>
```

When this is the case, there's a deserialization error that is swallowed and igor spends 10 seconds executing the retries of this call with the same result each time.  In `BuildController#getJobStatus` the git details are optional so the transaction still succeeds without that information but it takes so long now that other services calling igor may time out.

I'm not sure if there's a good way to handle multiple actions (in the cases we've seen they aren't even necessarily commits from the same repo) but it seems like it would be good to fail after the first deserialization error rather than retrying.


willgorman:
@spinnaker/reviewers Would it be possible to have someone take a look at this?

willgorman:
@tomaslin Do you think this looks ok?

willgorman:
@tomaslin @lwander checking back in on this, does this seem ok to merge?

PR #268, opened 2018-05-28
feat(*): Adding locking support to build monitors

robzienert:
Depends on https://github.com/spinnaker/kork/pull/150.

This will allow us to run more than one igor. Since this is new functionality, the LockManager defaults to completely disabled.


igor-web/src/main/groovy/com/netflix/spinnaker/igor/polling/CommonPollingMonitor.java
```diff
@@ -106,6 +115,21 @@ public PollContext getPollContext(String partition) {
...
     @Override
     public void pollSingle(PollContext ctx) {
+        if (lockManager.isPresent()) {
+            LockOptions lockOptions = new LockOptions()
+                .withLockName(format("%s.%s", getName(), ctx.partitionName))
+                .withLeaseDuration(Duration.ofMillis(getPollInterval()));
+
+            lockManager.get().acquireLock(lockOptions, () -> {
+                internalPollSingle(ctx);
+                return null;
```
robzienert: @jeyrschabu It'd be nice if there were a `Runnable` interface as well, for cases like this where we don't have a return value resulting from acquiring the lock.

PR #269, opened 2018-06-01
feat(*) Locking support

jeyrschabu:
- build monitors locking
- allows running multiple igor for polling


igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerRegistryCacheV2KeysMigration.java
```diff
@@ -62,7 +63,7 @@
     private final AtomicBoolean running = new AtomicBoolean();
 
     @Autowired
-    public DockerRegistryCacheV2KeysMigration(RedisClientDelegate redis,
+    public DockerRegistryCacheV2KeysMigration(@Qualifier("redisClientDelegate") RedisClientDelegate redis,
```
robzienert: Why is this qualifier necessary?

jeyrschabu: ha! had a case of two beans found and only one is needed failure when I first ran this. Verified later on that I don't need the qualifier. Just forgot to yank it out. Good catch sir!

PR #273, opened 2018-06-07
fix(jenkins) return multiple scms when jenkins has multiple scms and remoteUrl calls

kevinawoo:
- Some jenkins jobs will trigger with multiple scms, this however caused igor to break and not return any scm information
- Jenkins XML api requires remoteUrl to be pluralized


tomaslin:
lgtm

PR #276, opened 2018-06-08
fix(build): Fix spring warnings about validated classes

ezimanyi:
As of Spring Boot 1.5, `@ConfigurationProperties` classes that use constraint annotations should also be annotated with `@Validated`; this is currently throwing warnings but may actually break in the future.


ezimanyi:
https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-1.5-Release-Notes#configurationproperties-validation

PR #279, opened 2018-06-11
refactor(lock): Make locking more configurable

jeyrschabu:
- Added a config for lock properties

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/polling/LockService.java
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/polling/LockService.java
@@ -1,0 +1,34 @@
+package com.netflix.spinnaker.igor.polling;
+
+import com.netflix.spinnaker.kork.lock.LockManager;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import java.time.Duration;
+import java.util.Optional;
+
+public class LockService {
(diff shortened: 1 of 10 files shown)


jeyrschabu:
To enable locking:
```
locking:
  enabled: true
  maximumLockDurationMillis: 120000 // if defined it will override the maximum lock duration
  heartbeatRateMillis: 3000 // Optional
  leaseDurationMillis: 30000 // Optional
```

`Igor` uses the Poll Interval for as the maximum lock duration. This PR allows to override that value

igor-web/src/main/groovy/com/netflix/spinnaker/igor/polling/LockService.java
```diff
@@ -0,0 +1,42 @@
...
+
+@Component
+@ConditionalOnProperty("locking.enabled")
+public class LockService {
+    private final Logger log = LoggerFactory.getLogger(LockService.class);
+    private final LockManager delegate;
+    private Duration configuredMaxLockDuration = null;
+
+    @Autowired
+    public LockService(LockManagerConfigProperties lockManagerConfigProperties, LockManager redisLockManager) {
```
robzienert: It's not necessarily going to be a redis lock manager.

jeyrschabu:
Updated @robzienert PTAL

PR #281, opened 2018-06-13
fix(jenkins) return multiple scms when jenkins has multiple scms and remoteUrl calls

kevinawoo:
same as https://github.com/spinnaker/igor/pull/273, but for `release-1.8.x`


kevinawoo:
hm... spinnakerbot, not sure what you mean?
<kbd>

</kbd>

tomaslin:
lgtm

lwander:
Bad @spinnakerbot

PR #282, opened 2018-06-15
chore(build): Add debug flag to igor build

ezimanyi:
Starting igor with ./gradlew -DDEBUG=true now causes the JVM to listen on port 8188 for a remote debugger.


ezimanyi:
The diff on `README.md` looks big b/c my editor changed CRLF to LF (which I think fits with how most of our files are, so I didn't revert that); viewing with whitespace diff off recommended.

PR #285, opened 2018-06-20
fix(docker): Fix infinite loop in redis key migration

ezimanyi:


ezimanyi:
@jtk54 : This might be related to the issues you were seeing with load on redis

robzienert:
Oops

lwander:
Can we patch this into 1.8?

ezimanyi:
@lwander : Yup, I'll make a PR to puatch into 1.8.

PR #289, opened 2018-06-25
feat(echo): Adding a health indicator for echo

jeyrschabu:
- validates connectivity to echo
- must pass at least once before Igor goes UP
- up by default if echo is not configured

### DIFF
--- a/igor-web/src/main/java/com/netflix/spinnaker/igor/health/EchoServiceHealthIndicator.java
+++ b/igor-web/src/main/java/com/netflix/spinnaker/igor/health/EchoServiceHealthIndicator.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright 2018 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the 'License')
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an 'AS IS' BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.igor.health;
+
+import com.netflix.spectator.api.Registry;
+import com.netflix.spectator.api.patterns.PolledMeter;
+import com.netflix.spinnaker.igor.build.model.GenericBuild;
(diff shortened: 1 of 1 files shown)


igor-web/src/main/java/com/netflix/spinnaker/igor/health/EchoServiceHealthIndicator.java
```diff
@@ -0,0 +1,88 @@
...
+    @Scheduled(fixedDelay = 120000L)
+    void checkHealth() {
+        echoService.ifPresent(s -> {
+            try {
+                s.postEvent(event);
+                upOnce.set(true);
+                errors.set(0);
+            } catch (Exception e) {
+                errors.set(1);
+                log.error("Unable to connect to Echo", e);
```
ajordens: might be worth capturing the most recent exception and throwing it from the `health()` -- benefit would be that hitting `/health` would show an exception stacktrace and you wouldn't necessary need to dig into logs.

igor-web/src/main/java/com/netflix/spinnaker/igor/health/EchoServiceHealthIndicator.java
```diff
@@ -0,0 +1,88 @@
...
+    private final Optional<EchoService> echoService;
+    private final AtomicLong errors;
+
+    static final GenericBuildEvent event = buildGenericEvent();
+
+    @Autowired
+    EchoServiceHealthIndicator(Registry registry,
+                               EchoService echoService) {
+        this.echoService = Optional.ofNullable(echoService);
+        this.upOnce = !this.echoService.isPresent() ? new AtomicBoolean(true) : new AtomicBoolean(false);
```
ajordens: could also make this `HealthIndicator` condition on echo being enabled

igor-web/src/main/java/com/netflix/spinnaker/igor/health/EchoServiceHealthIndicator.java
```diff
@@ -0,0 +1,88 @@
...
+    private Logger log = LoggerFactory.getLogger(EchoServiceHealthIndicator.class);
+    private final AtomicBoolean upOnce;
+    private final Optional<EchoService> echoService;
+    private final AtomicLong errors;
+
+    static final GenericBuildEvent event = buildGenericEvent();
+
+    @Autowired
+    EchoServiceHealthIndicator(Registry registry,
+                               EchoService echoService) {
```
ajordens: If echo is disabled, are you actually going to get a null `echoService`? I thought it would have failed to launch saying it couldn't find a bean (in other spots we've done the equivalent of `Optional<EchoService> echoService` as the constructor parameter to get around this.

PR #293, opened 2018-06-27
refactor(healthcheck): Improving Poll Monitor health check

jeyrschabu:
- Only recalculate Poller's health if the node has been UP at least once and there are DOWN pollers.
- adding a metric for down pollers


igor-web/src/main/java/com/netflix/spinnaker/igor/health/PollingMonitorHealth.java
```diff
@@ -0,0 +1,130 @@
...
+
+    @Override
+    public Health health() {
+        final Collection<PollingMonitor> pollers = getPollingMonitors();
+        List<PollingMonitor> downPollingMonitors = pollers
+            .stream()
+            .filter(hasBeenFiveTimesPollIntervalSinceLastPollPredicate())
+            .collect(toList());
+
+        if (upOnce.get() && !downPollingMonitors.isEmpty()) {
```
cfieber: This seems like it is going to initially fall through to returning UP before upOnce has been set to true

Not sure we want that since I think we want to make sure we have had one successful poll of each before we initially go into service

igor-web/src/test/groovy/com/netflix/spinnaker/igor/health/PollingMonitorHealthSpec.groovy
```diff
@@ -0,0 +1,78 @@
...
+            getPollInterval() >> 60 //seconds
+            getLastPoll() >> dockerLastPoll
+            isInService() >> true
+            getName() >> "dockerMonitor"
+        }
+
+        applicationContext.getBeansOfType(PollingMonitor) >> [jenkins: jenkinsBuildMonitor, docker: dockerMonitor]
+
+
+        when:
```
cfieber: could rewrite this when, then block as

```
expect:
pollingMonitorHealth.health().status == healthStatus
```

igor-web/src/main/java/com/netflix/spinnaker/igor/health/PollingMonitorHealth.java
jeyrschabu: Chatted with @cfieber offline about the circular dependency between poller's health and discovery status. This fixes that.
The plan is to add some additional health checks per poller and that will come in another PR.

PR #294, opened 2018-07-02
fix(health): new install of igor has no pollers, but it's still healthy

kevinawoo:
When igor is first booted up, it's "unhealthy" because there's no pollers configured, however igor is actually healthy, awaiting configurations.

Note: I'm not sure how to handle health in this scenario though:
- igor has been configured with Jenkins and Docker
- a new version of igor is deployed with messed up configs so that Jenkins and Docker yml stanzas are gone
- what state is igor in now?


lwander:
What makes that scenario different from any other "we pushed bad spinnaker config" scenario?

robzienert:
Happy to merge this as-is, but will need conflicts resolved.

kevinawoo:
I believe what's in master address it. Closing for now.

PR #297, opened 2018-07-10
feat(wercker): add wercker CI

shihchang:
Add Wercker as a  CI provider.
- WerckerBuildMonitor polls Wercker runs for each wercker master
- WerckerCache caches last polling time and generates buildNumber for each runId as wercker does not have a buildNumber for each run.
- InfoController added an 'type' parameter for getting the build masters.

Related PRs:
https://github.com/spinnaker/halyard/pull/960


shihchang:
@markxnelson

shihchang:
@spinnaker/reviewers

related PRs:
spinnaker/halyard#960
spinnaker/deck#5519
spinnaker/echo#295
spinnaker/gate#574

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
```diff
@@ -1,5 +1,6 @@
 /*
  * Copyright 2015 Netflix, Inc.
+ * Copyright (c) 2017, 2018, Oracle Corporation and/or its affiliates. All rights reserved.
```
skim1420: [No need](http://www.apache.org/dev/apply-license.html#contributor-copyright) to append Copyrights on these existing files.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
```diff
@@ -0,0 +1,360 @@
...
+				runs.add(run);
+			} else {
+				runs = [run];
+				pipelineRuns.put(pipelineKey, runs);
+			}
+		});
+		return pipelineRuns;
+	}
+	
+	//TODO this is Jenkins JobConfig
```
sbwsg: This `todo` doesn't describe what needs to be done.  Please add a little more context here.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/model/Application.groovy
```diff
@@ -0,0 +1,45 @@
...
+ * You can obtain one at https://www.apache.org/licenses/LICENSE-2.0.html
+ */
+package com.netflix.spinnaker.igor.wercker.model
+
+/**
+ * Represents a Wercker application
+ */
+@groovy.transform.ToString
+class Application {
+    /*    {
```
sbwsg: Is this a debug comment or did you intend to leave it here?  Please remove if it's an artifact of debugging.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/model/Owner.groovy
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * The contents of this file are subject to the Apache License Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * If a copy of the Apache License Version 2.0 was not distributed with this file,
+ * You can obtain one at https://www.apache.org/licenses/LICENSE-2.0.html
+ */
+package com.netflix.spinnaker.igor.wercker.model
+
+class Owner {
+    /*"owner": {
```
sbwsg: Same as Application, please remove if no longer necessary.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/model/Pipeline.groovy
```diff
@@ -0,0 +1,34 @@
...
+class Pipeline {
+    String id
+    String url
+    String name
+    String permissions
+    String pipelineName //In the wercker.yml, I think
+    boolean setScmProviderStatus
+    String type
+	Application application
+    /*
```
sbwsg: Same as above, please remove if no longer necessary.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/model/Run.groovy
```diff
@@ -0,0 +1,86 @@
...
+            "permissions": "read",
+            "pipelineName": "deploy-kube",
+            "setScmProviderStatus": false,
+            "type": "pipeline"
+        }
+     */
+
+	
+	
+//		public Instant getFinishedAt() {
```
sbwsg: Remove dead code.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/model/Run.groovy
```diff
@@ -0,0 +1,86 @@
...
+    int progress
+    String result
+    String status
+	
+    Owner user
+    String pipelineId
+    Pipeline pipeline
+    Application application
+
+    /*
```
sbwsg: Remove if no longer needed.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/model/Run.groovy
```diff
@@ -0,0 +1,86 @@
...
+//		public Instant getFinishedAt() {
+//			return finishedAt;
+//		}
+//	
+//		public void setFinishedAt(Instant finishedAt) {
+//			this.finishedAt = finishedAt;
+//		}
+	
+	public String toString() {
+		return "Run(" +id+", finishedAt=" + finishedAt+"," + application?.name + "/"+pipeline?.id+")";
```
sbwsg: Check leading whitespace throughout.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/model/Workflow.groovy
```diff
@@ -0,0 +1,101 @@
...
+ * you may not use this file except in compliance with the License.
+ * If a copy of the Apache License Version 2.0 was not distributed with this file,
+ * You can obtain one at https://www.apache.org/licenses/LICENSE-2.0.html
+ */
+package com.netflix.spinnaker.igor.wercker.model
+
+class Workflow {
+    String url
+    WorkflowData data
+    /*
```
sbwsg: Remove if no longer needed.  Same for JSON object on line 89.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/docker/DockerRegistryCacheV2KeysMigrationSpec.groovy
```diff
@@ -59,7 +59,10 @@ class DockerRegistryCacheV2KeysMigrationSpec extends Specification {
             keyFactory, igorConfigurationProperties, Schedulers.immediate())
     }
 
-    def tearDown() {
+    void cleanup() {
```
sbwsg: Why was this renamed?  I'm not familiar with Igor's testing.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSchedulingSpec.groovy
```diff
@@ -0,0 +1,141 @@
...
+import com.netflix.spectator.api.NoopRegistry
+import com.netflix.spinnaker.igor.IgorConfigurationProperties
+import com.netflix.spinnaker.igor.config.WerckerProperties
+import com.netflix.spinnaker.igor.history.EchoService
+import com.netflix.spinnaker.igor.wercker.WerckerService
+import com.netflix.spinnaker.igor.wercker.model.Run
+import com.netflix.spinnaker.igor.model.BuildServiceProvider
+import com.netflix.spinnaker.igor.service.BuildMasters
+import com.netflix.spinnaker.kork.eureka.RemoteStatusChangedEvent
+
```
sbwsg: Remove newline.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSchedulingSpec.groovy
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * Ensures that build monitor runs periodically
+ */
+class WerckerBuildMonitorSchedulingSpec extends Specification {
+
+    WerckerCache cache = Mock(WerckerCache)
+    WerckerService werckerService = Mock(WerckerService)
+    WerckerBuildMonitor monitor
+	Optional<EchoService> echoService = Optional.of(Mock(EchoService))
```
sbwsg: Check leading whitespace throughout.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSpec.groovy
```diff
@@ -0,0 +1,258 @@
...
+import com.netflix.spinnaker.igor.wercker.model.Application
+import com.netflix.spinnaker.igor.wercker.model.Pipeline
+import com.netflix.spinnaker.igor.wercker.model.Run
+import com.netflix.spinnaker.igor.wercker.model.Owner
+import com.netflix.spinnaker.igor.model.BuildServiceProvider
+import com.netflix.spinnaker.igor.service.BuildMasters
+import com.netflix.spinnaker.kork.eureka.RemoteStatusChangedEvent
+import com.netflix.spinnaker.igor.history.model.Event
+import com.netflix.spinnaker.igor.config.WerckerProperties.WerckerHost
+
```
sbwsg: Remove newline and check sort order.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSpec.groovy
```diff
@@ -0,0 +1,258 @@
...
+import spock.lang.Specification
+import spock.lang.Subject
+
+import java.util.Optional
+import java.util.concurrent.TimeUnit
+import java.util.List
+
+import org.slf4j.Logger
+import retrofit.RetrofitError
+import rx.schedulers.Schedulers
```
sbwsg: Move these imports up to the same group as `rx.schedulers.TestScheduler` & check sort order.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSpec.groovy
```diff
@@ -0,0 +1,258 @@
...
+
+class WerckerBuildMonitorSpec extends Specification {
+
+    WerckerCache cache = Mock(WerckerCache)
+    WerckerBuildMonitor monitor
+	EchoService echoService = Mock(EchoService)
+	WerckerClient client
+    WerckerService mockService = Mock(WerckerService)
+    WerckerService werckerService
+	String werckerDev = 'https://dev.wercker.com/'
```
sbwsg: Check leading whitespace throughout.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerCacheSpec.groovy
```diff
@@ -0,0 +1,112 @@
...
+ * If a copy of the Apache License Version 2.0 was not distributed with this file,
+ * You can obtain one at https://www.apache.org/licenses/LICENSE-2.0.html
+ */
+package com.netflix.spinnaker.igor.wercker
+
+import com.netflix.spinnaker.igor.IgorConfigurationProperties
+import com.netflix.spinnaker.kork.jedis.EmbeddedRedis
+import com.netflix.spinnaker.kork.jedis.JedisClientDelegate
+import com.netflix.spinnaker.kork.jedis.RedisClientDelegate
+import com.netflix.spinnaker.igor.wercker.model.Run;
```
sbwsg: Check import sort order.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerCacheSpec.groovy
```diff
@@ -0,0 +1,112 @@
...
+	
+    EmbeddedRedis embeddedRedis = EmbeddedRedis.embed()
+    RedisClientDelegate redisClientDelegate = new JedisClientDelegate(embeddedRedis.pool as JedisPool)
+
+    @Subject
+    WerckerCache cache = new WerckerCache(redisClientDelegate, new IgorConfigurationProperties())
+
+    final master = 'testWerckerMaster'
+    final test = 'test'
+	final pipeline = 'myOrg/myApp/myTestPipeline'
```
sbwsg: Check leading whitespace throughout.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerClientSpec.groovy
```diff
@@ -0,0 +1,111 @@
...
+package com.netflix.spinnaker.igor.wercker
+
+import com.netflix.spinnaker.igor.config.*
+import com.netflix.spinnaker.igor.config.WerckerProperties.WerckerHost
+import com.netflix.spinnaker.igor.wercker.model.*
+import com.squareup.okhttp.mockwebserver.MockResponse
+import com.squareup.okhttp.mockwebserver.MockWebServer
+import spock.lang.Shared
+import spock.lang.Specification
+import java.util.List
```
sbwsg: Separate java imports and group non-java imports.  Check sort order.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerClientSpec.groovy
```diff
@@ -0,0 +1,111 @@
...
+        server = new MockWebServer()
+    }
+
+    void cleanup() {
+        server.shutdown()
+    }
+
+    void 'get all applications'() {
+        setup:
+		def authHeader = "my_authHeader"
```
sbwsg: Check leading whitespace throughout.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerServiceSpec.groovy
```diff
@@ -0,0 +1,159 @@
...
+import spock.lang.Shared
+import spock.lang.Specification
+import spock.lang.Unroll
+
+class WerckerServiceSpec extends Specification {
+
+	WerckerCache cache = Mock(WerckerCache)
+
+	@Shared
+	WerckerClient client
```
sbwsg: Check leading whitespace throughout.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/Front50Config.groovy
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright 2014 Netflix, Inc.
```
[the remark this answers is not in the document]
shihchang: This Front50Config and the Front50Service files are replicated from some existing files with "Copyright 2014 Netflix, Inc." so l leave it instead of replacing it to the Oracle one. And I will append another line of Oracle copyright.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/WerckerConfig.groovy
```diff
@@ -0,0 +1,65 @@
...
+import com.netflix.spinnaker.igor.wercker.WerckerClient
+import com.netflix.spinnaker.igor.wercker.WerckerService
+import com.squareup.okhttp.OkHttpClient
+import groovy.transform.CompileStatic
+import groovy.util.logging.Slf4j
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
+import org.springframework.boot.context.properties.EnableConfigurationProperties
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import com.netflix.spinnaker.retrofit.Slf4jRetrofitLogger
```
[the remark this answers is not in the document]
shihchang: I am reformatting the new codes with the 'JAVA conventions' eclipse code formatter and 2 space indents as described in https://www.spinnaker.io/community/contributing/back-end-code/. Or does Spinnaker has a eclipse code formatter?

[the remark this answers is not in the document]
sbwsg: I'd suggest reformatting by following the example of other code in the repo.  Every other groovy file in Igor appears to use 4 spaces for its indents or are you seeing something different?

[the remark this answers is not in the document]
sbwsg: Unfortunately I'm not familiar enough with eclipse to advise on setting up code formatting for Spinnaker.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/docker/DockerRegistryCacheV2KeysMigrationSpec.groovy
shihchang: it should have been cleanup not tearDown. The tearDown never got called and leak the the redis in memory. After several runs of "gradlew clean build", my laptop memory was full of redis because of these.

sbwsg: Ah, good to know, thanks!

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/WerckerConfig.groovy
```diff
@@ -0,0 +1,65 @@
...
+            WerckerProperties.WerckerHost host ->
+            log.debug "bootstrapping Wercker ${host.address} as ${host.name}"
+            [(host.name): new WerckerService(host, cache, werckerClient(host))]
+        })
+
+        buildMasters.map.putAll werckerMasters
+        werckerMasters
+    }
+
+    static WerckerClient werckerClient(WerckerHost host, int timeout = 30000) {
```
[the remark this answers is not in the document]
shihchang: I changed it to use igorConfigurationProperties.client.timeout as in JenkinsConfig

ezimanyi:
@sbwsg : I didn't get a chance to take a look today, as I had a few other things on my plate, but will take a closer look tomorrow.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
```diff
@@ -0,0 +1,349 @@
...
+     * Returns List of all Wercker jobs in the format of type/org/app/pipeline 
+     */
+    public List<String> getJobs() {
+        List<String> jobs = []
+        long start = System.currentTimeMillis();
+        List<Application> apps = getApplications();
+        start = System.currentTimeMillis()
+        apps.each { app ->
+            try {
+                List<Pipeline> pipelines = app.pipelines? app.pipelines : []; //getPipelines(app.owner.name, app.name);
```
ezimanyi: In Groovy this can be just `app.pipelines ?: []` (the "Elvis operator")

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
```diff
@@ -0,0 +1,349 @@
...
+            } catch(retrofit.RetrofitError err) {
+                log.error "Error getting pipelines for ${app.owner.name } ${app.name} ${err}"
+            }
+        }
+        log.debug "getPipelines: ${jobs.size()} pipelines in ${System.currentTimeMillis() - start}ms"
+        return jobs
+    }
+
+    List<Application> getApplications() {
+        new SimpleHystrixCommand<List<Application>>(groupKey, buildCommandKey("getApplications"), {
```
ezimanyi: I'm curious why this (and the next) call to the `werckerClient` go through `Hystrix` but none of the other ones do.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
```diff
@@ -0,0 +1,349 @@
...
+
+    /**
+     * A CommandKey should be unique per group (to ensure broken circuits do not span Wercker masters)
+     */
+    private String buildCommandKey(String id) {
+        return "${groupKey}-${id}"
+    }
+
+    List<Run> getBuilds(String appAndPipelineName) {
+        String[] split = appAndPipelineName.split(SPLITOR)
```
ezimanyi: I think most of this function can be replaced by a call to `getPipelineId`

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
```diff
@@ -0,0 +1,349 @@
...
+            } catch(retrofit.RetrofitError err) {
+                log.info "Error getting pipelines for ${owner} ${appName} ${err} ${err.getClass()}"
+            }
+        }
+        return name;
+    }
+
+    Map<String, List<Run>> getRunsSince(Set<String> pipelines, long since) {
+        long start = System.currentTimeMillis();
+        List<String> pipelineIds = pipelines.collect { getPipelineId(it) };
```
ezimanyi: Variable is unused

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
```diff
@@ -0,0 +1,349 @@
...
+        Map<String, String> pipelineKeys = pipelines.collectEntries { [(getPipelineId(it)) : (it)] };
+        List<String> pids = pipelineKeys.keySet().asList();
+        Map<String, List<Run>> pipelineRuns = [:];
+        List<Run> allRuns = werckerClient.getRunsSince(authHeaderValue, branch, pids, limit, since);
+        log.debug "getRunsSince for pipelines:${pipelines} : ${allRuns.size()} runs in ${System.currentTimeMillis() - start}ms!!"
+        allRuns.forEach({ run ->
```
ezimanyi: Just a minor style point, but I think this block can be replaced with a Groovy `.groupBy`

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
```diff
@@ -0,0 +1,349 @@
...
+                runs.add(run);
+            } else {
+                runs = [run];
+                pipelineRuns.put(pipelineKey, runs);
+            }
+        });
+        return pipelineRuns;
+    }
+
+    Map<String, List<Run>> getRunsSince(long since) {
```
ezimanyi: This is mostly a duplicate of the `getRunsSince` that accepts `pipelines`.  Can the common logic be factored out?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
```diff
@@ -0,0 +1,349 @@
...
+        List<String> pids = pipelineKeys.keySet().asList();
+        Map<String, List<Run>> pipelineRuns = [:];
+        List<Run> allRuns = werckerClient.getRunsSince(authHeaderValue, branch, pids, limit, since);
+        log.debug "getRunsSince for pipelines:${pipelines} : ${allRuns.size()} runs in ${System.currentTimeMillis() - start}ms!!"
+        allRuns.forEach({ run ->
+            String pipelineKey = pipelineKey(run, pipelineKeys);
+            run.startedAt = run.startedAt?:run.createdAt;
+            List<Run> runs = pipelineRuns.get(pipelineKey);
+            if (runs) {
+                runs.add(run);
```
ezimanyi: Some un-necessary semicolons here

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitor.groovy
```diff
@@ -0,0 +1,335 @@
...
+                    jobs.add(stage['job'])
+                }
+            })
+        })
+        return jobs
+    }
+
+    static Comparator<Run> finishedAtComparator = new Comparator<Run>() {
+        @Override
+        public int compare(Run r1, Run r2) {
```
ezimanyi: The `finishedAtComparator` and `startedAtComparator` are almost duplicated---it might be easier to use something in the `Comparator` library rather than implement this manually.

Something like `Comparator.nullsLast({ r -> r.finishedAt?.fastTime })` should be equivalent and would make the intent more clear.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitor.groovy
```diff
@@ -0,0 +1,335 @@
...
+        return Collections.max(runs, startedAtComparator)
+    }
+
+    /**
+     * wercker.pipeline = project|job
+     * wercker.run = build
+     */
+    private void processRuns( WerckerService werckerService, String master, String pipeline,
+            List<PipelineDelta> delta, List<Run> runs) {
+        List<Run> allRuns = runs? runs : werckerService.getBuilds(pipeline)
```
ezimanyi: `runs ?: werkerService.getBuilds(pipeline)` is more groovy-like

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitor.groovy
```diff
@@ -0,0 +1,335 @@
...
+            log.error("Error processing runs for [{}:{}]", kv("master", master), kv("pipeline", pipeline), e)
+            if (e.cause instanceof RetrofitError) {
+                def re = (RetrofitError) e.cause
+                log.error("Error communicating with Wercker for [{}:{}]: {}", kv("master", master), kv("pipeline", pipeline), kv("url", re.url), re)
+            }
+        }
+    }
+
+    private List<Run> onlyInLookBackWindow(List<Run> builds) {
+        use(TimeCategory) {
```
ezimanyi: Nice, had not seen `TimeCategory` before

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitor.groovy
```diff
@@ -0,0 +1,335 @@
...
+    protected void commitDelta(PipelinePollingDelta delta, boolean sendEvents) {
+        String master = delta.master
+        delta.items.parallelStream().forEach { pipeline ->
+            //job = Wercker pipeline (org/app/pipeline)
+            // post event for latest finished run
+            pipeline.completedBuilds.forEach { run ->
+                //build = Wercker run
+                Boolean eventPosted = cache.getEventPosted(master, pipeline.name, run.id)
+                GenericBuild build = toBuild(master, pipeline.name, run)
+                if (!eventPosted && echoService.isPresent() && sendEvents) {
```
ezimanyi: Is there a reason to gate this on `echoService.isPresent()`; this is already checked in `postEvent` and throws an error if Echo isn't present (which seems desirable since this won't work without Echo).

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerCache.java
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    static Run useCreatedAtIfNotStarted(Run run) {
+        if (run.getStartedAt() == null) {
+            run.setStartedAt(run.getCreatedAt());
+        }
+        return run;
+    }
+
+    static Comparator<Run> runStartedAtComparator = new Comparator<Run>() {
```
ezimanyi: Similar to my above comment, it should be possible to greatly simplify this using functions in the `Comparator` library.

Something like `Comparator.nullsLast({ r -> (r.getStartedAt() ?: r.getCreatedAt())?.toInstant().toEpochMilli() })`.  That would also eliminate the need for `useCreatedIfNotStarted` which has the effect of mutating the objects being compared.  Generally I'd expect a comparator to have no side effects.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerCache.java
```diff
@@ -0,0 +1,210 @@
...
+        redisClientDelegate.withCommandsClient(c -> {
+            c.del(makeKey(master, job));
+        });
+    }
+
+    private String makeEventsKey(String master, String job) {
+        return makeKey(master, job) + ":" + POLL_STAMP + ":events";
+    }
+
+    private String makeKey(String master, String job) {
```
ezimanyi: This looks to be the same key naming convention as Jenkins.  Are we sure we won't see conflicts for a user running both Jenkins and Wercker, if they have jobs with the same (or similar) names?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/model/Application.groovy
```diff
@@ -0,0 +1,22 @@
...
+ * You can obtain one at https://www.apache.org/licenses/LICENSE-2.0.html
+ */
+package com.netflix.spinnaker.igor.wercker.model
+
+/**
+ * Represents a Wercker application
+ */
+@groovy.transform.ToString
+class Application {
+    String id;
```
ezimanyi: Semicolons unnecessary in this file

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/model/Application.groovy
```diff
@@ -0,0 +1,22 @@
...
+ * you may not use this file except in compliance with the License.
+ * If a copy of the Apache License Version 2.0 was not distributed with this file,
+ * You can obtain one at https://www.apache.org/licenses/LICENSE-2.0.html
+ */
+package com.netflix.spinnaker.igor.wercker.model
+
+/**
+ * Represents a Wercker application
+ */
+@groovy.transform.ToString
```
ezimanyi: No need to qualify `ToString` here, can just do `@ToString`

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/model/Run.groovy
```diff
@@ -0,0 +1,37 @@
...
+    String result
+    String status
+
+    Owner user
+    String pipelineId
+    Pipeline pipeline
+    Application application
+
+    public String toString() {
+        return "Run(" +id+", finishedAt=" + finishedAt+"," + application?.name + "/"+pipeline?.id+")";
```
ezimanyi: Would me more readable using either variable interpolation, or `sprintf`

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/model/Pipeline.groovy
```diff
@@ -0,0 +1,24 @@
...
+ * you may not use this file except in compliance with the License.
+ * If a copy of the Apache License Version 2.0 was not distributed with this file,
+ * You can obtain one at https://www.apache.org/licenses/LICENSE-2.0.html
+ */
+package com.netflix.spinnaker.igor.wercker.model
+
+/**
+ * Represents a Wercker Pipeline
+ */
+@groovy.transform.ToString
```
ezimanyi: Can just be `@ToString`

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/model/Run.groovy
```diff
@@ -0,0 +1,37 @@
...
+ * you may not use this file except in compliance with the License.
+ * If a copy of the Apache License Version 2.0 was not distributed with this file,
+ * You can obtain one at https://www.apache.org/licenses/LICENSE-2.0.html
+ */
+package com.netflix.spinnaker.igor.wercker.model
+
+/**
+ * Represents a Wercker Run
+ */
+@groovy.transform.ToString
```
ezimanyi: Can just be `@ToString`

igor-web/src/test/groovy/com/netflix/spinnaker/igor/build/InfoControllerSpec.groovy
```diff
@@ -142,6 +142,7 @@ class InfoControllerSpec extends Specification {
 
         then:
         1 * buildMasters.filteredMap(BuildServiceProvider.JENKINS) >> ["travis-master1": []]
+        1 * buildMasters.filteredMap(BuildServiceProvider.WERCKER) >> ["travis-master1": []]
```
ezimanyi: I'm not sure I understand this test case---why are we adding "travis-master1" to the Wercker masters in the case that's testing Travis?  I think this test case should be exercising the Travis code, and a new test case should exercise the Wercker code.

(Also, this is pre-existing, but I'm not sure why the above line adds the master to Jenkins, not Travis.)

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSpec.groovy
```diff
@@ -0,0 +1,251 @@
...
+
+    void setup() {
+        client = Mock(WerckerClient)
+        werckerService = new WerckerService(
+                new WerckerHost(name: master, address: werckerDev), cache, client)
+    }
+
+    final MASTER = 'MASTER'
+    final pipeline = 'myOrg/myApp/pollingTest'
+    final List<String> APPS = []
```
ezimanyi: This looks unused

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSpec.groovy
```diff
@@ -0,0 +1,251 @@
...
+        then: 'initial poll'
+        1 * buildMasters.filteredMap(BuildServiceProvider.WERCKER) >> [MASTER: mockService]
+        1 * buildMasters.map >> [MASTER: mockService]
+        1 * echoService.postEvent(_)
+
+        cleanup:
+        monitor.stop()
+    }
+
+    void 'select latest one form multiple completed runs'() {
```
ezimanyi: s/form/from/

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSpec.groovy
```diff
@@ -0,0 +1,251 @@
...
+        mockService.getRunsSince(_) >> [pipeline: runs1]
+        cache.getBuildNumber(_, _, _) >> 1
+
+        when:
+        monitor.onApplicationEvent(Mock(RemoteStatusChangedEvent))
+        scheduler.advanceTimeBy(1L, TimeUnit.SECONDS.MILLISECONDS)
+
+        then: 'initial poll'
+        1 * buildMasters.filteredMap(BuildServiceProvider.WERCKER) >> [MASTER: mockService]
+        1 * buildMasters.map >> [MASTER: mockService]
```
ezimanyi: This looks to be duplicating the mock in the `given` section.  If we're stubbing (and don't really care about testing the number of calls) we can move it to `given`; if testing the number of calls is part of the test then it should stay in `then`.  (My guess is that the number of calls to `buildMasters` is an implementation detail so it should go in `given`)

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSpec.groovy
```diff
@@ -0,0 +1,251 @@
...
+    }
+
+    final MASTER = 'MASTER'
+    final pipeline = 'myOrg/myApp/pollingTest'
+    final List<String> APPS = []
+    final TestScheduler scheduler = new TestScheduler()
+
+    void 'no finished run'() {
+        given:
+        cache.getJobNames(MASTER) >> ['pipeline']
```
ezimanyi: Most of this setup code is duplicated with the next test (and appears on a number of them).  It would probably make sense to pull this into a function, or just have multiple tests under the same `given`.  I think all the tests (except maybe the last one with multiple pipelines) could just have one setup and a data table (or multiple when/then).

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSpec.groovy
```diff
@@ -0,0 +1,251 @@
...
+        ]
+        cache.getLastPollCycleTimestamp(_, _) >> (now - 1000)
+        mockService.getRunsSince(_) >> [pipeline: runs1]
+        cache.getBuildNumber(_, _, _) >> 1
+        scheduler.advanceTimeBy(2L, TimeUnit.SECONDS.MILLISECONDS)
+
+        then:
+        1 * buildMasters.filteredMap(BuildServiceProvider.WERCKER) >> [MASTER: mockService]
+        1 * buildMasters.map >> [MASTER: mockService]
+        1 * echoService.postEvent(_)
```
ezimanyi: Are we actually testing that we pick the latest from multiple runs, as the description suggests?  It looks like we're just testing that a single event was posted.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSpec.groovy
```diff
@@ -0,0 +1,251 @@
...
+        monitor.stop()
+    }
+
+    void 'get runs of multiple pipelines'() {
+        setup:
+        BuildMasters buildMasters = Mock(BuildMasters)
+        buildMasters.map >> [MASTER: werckerService]
+        monitor = monitor(buildMasters)
+        monitor.worker = scheduler.createWorker()
+        cache.getBuildNumber(_, _, _) >> 1
```
ezimanyi: This, and the next line, might be more maintainable with a wildcard `cache.getBuildNumber(*_)` so that we don't need to add an underscore if an argument is ever added to `getBuildNumber`

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSchedulingSpec.groovy
```diff
@@ -0,0 +1,137 @@
...
+ */
+class WerckerBuildMonitorSchedulingSpec extends Specification {
+
+    WerckerCache cache = Mock(WerckerCache)
+    WerckerService werckerService = Mock(WerckerService)
+    WerckerBuildMonitor monitor
+    Optional<EchoService> echoService = Optional.of(Mock(EchoService))
+
+    final MASTER = 'MASTER'
+    final List<String> APPS = []
```
ezimanyi: Looks unused

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSchedulingSpec.groovy
```diff
@@ -0,0 +1,137 @@
...
+
+/**
+ * Ensures that build monitor runs periodically
+ */
+class WerckerBuildMonitorSchedulingSpec extends Specification {
+
+    WerckerCache cache = Mock(WerckerCache)
+    WerckerService werckerService = Mock(WerckerService)
+    WerckerBuildMonitor monitor
+    Optional<EchoService> echoService = Optional.of(Mock(EchoService))
```
ezimanyi: Looks unused

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSchedulingSpec.groovy
```diff
@@ -0,0 +1,137 @@
...
+
+        when: 'poll at 1 second'
+        scheduler.advanceTimeBy(2L, TimeUnit.SECONDS.MILLISECONDS)
+
+        then:
+        1 * buildMasters.filteredMap(BuildServiceProvider.WERCKER) >> [MASTER: werckerService]
+        1 * buildMasters.map >> [MASTER: werckerService]
+        1 * werckerService.getRunsSince(_) >> [:]
+
+        when: 'poll at 2 and 3 second'
```
ezimanyi: Is the name of this test correct?  Looks like we're advancing time from 1s to 5s, and observing 4 events.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSchedulingSpec.groovy
```diff
@@ -0,0 +1,137 @@
...
+        then: 'initial poll'
+        0 * buildMasters.filteredMap
+
+        when:
+        scheduler.advanceTimeBy(998L, TimeUnit.SECONDS.MILLISECONDS)
+
+        then:
+        0 * buildMasters.filteredMap
+
+        when: 'poll at 1 second'
```
ezimanyi: Minor, but maybe should change this to "no poll at 1 second" or something similar, as we're expecting the absence of polling.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
```diff
@@ -192,13 +196,21 @@ class BuildController {
             def queuedLocation = locationHeader.value
 
             queuedLocation.split('/')[-1]
-        } else if (buildMasters.filteredMap(BuildServiceProvider.TRAVIS).containsKey(master)) {
+        } else if (isTravis(master) || isWercker(master)) {
```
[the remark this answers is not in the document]
desagar: No there is not....that's an oversight. fixed.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
```diff
@@ -0,0 +1,349 @@
...
+    String authHeaderValue
+    String address
+    String master
+    WerckerCache cache
+
+    private static String SPLITOR = "/";
+    private static String branch = 'master'
+    private static limit = 300
+
+    public WerckerService(WerckerHost wercker, WerckerCache cache, WerckerClient werckerClient) {
```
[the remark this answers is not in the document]
desagar: removed all public qualifiers

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
```diff
@@ -0,0 +1,349 @@
...
+        this.address = wercker.address
+        this.setToken(wercker.token)
+    }
+
+    /**
+     * Custom setter for token, in order to re-set the authHeaderValue
+     * @param token
+     * @return
+     */
+    public setToken(String token) {
```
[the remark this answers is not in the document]
desagar: removed

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
```diff
@@ -0,0 +1,349 @@
...
+    }
+
+    @Override
+    GenericBuild getGenericBuild(final String job, final int buildNumber) {
+        GenericBuild genericBuild = new GenericBuild()
+        genericBuild.name = job
+        genericBuild.building = true
+        genericBuild.fullDisplayName = "Wercker Job " + job + " [" + buildNumber + "]"
+        genericBuild.number = buildNumber
+        String[] split = job.split(SPLITOR)
```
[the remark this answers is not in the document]
desagar: agreed - created a class QualifiedPipelineName to hold owner/app/pipeline and do the parsing/concatenation

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
```diff
@@ -0,0 +1,349 @@
...
+
+    @Override
+    List<GenericGitRevision> getGenericGitRevisions(final String job, final int buildNumber) {
+        return null
+    }
+
+    @Override
+    GenericBuild getGenericBuild(final String job, final int buildNumber) {
+        GenericBuild genericBuild = new GenericBuild()
+        genericBuild.name = job
```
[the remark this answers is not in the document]
desagar: done

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
```diff
@@ -0,0 +1,349 @@
...
+        } else {
+            throw new BuildController.InvalidJobParameterException(
+            "Could not retrieve pipeline ${pipelineName} for application ${appName} from Wercker!")
+        }
+    }
+
+    /**
+     * Returns List of all Wercker jobs in the format of type/org/app/pipeline 
+     */
+    public List<String> getJobs() {
```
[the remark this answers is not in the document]
desagar: removed

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
```diff
@@ -0,0 +1,349 @@
...
+        List<String> jobs = []
+        long start = System.currentTimeMillis();
+        List<Application> apps = getApplications();
+        start = System.currentTimeMillis()
```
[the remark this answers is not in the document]
desagar: fixed

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
```diff
@@ -0,0 +1,349 @@
...
+                break
+            case "failed":
+                return Result.FAILURE
+                break
+        }
+        return Result.UNSTABLE
+    }
+
+    Response stopRunningBuild (String appAndPipelineName, Integer buildNumber){
+        String[] split = appAndPipelineName.split(SPLITOR)
```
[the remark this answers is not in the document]
desagar: removed unused code

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
desagar: done

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
desagar: you are correct, fixed

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
desagar: fixed

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
desagar: nice Groovyism suggestion - thx :) done

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
desagar: done - moved it to groupRunsByPipeline

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
desagar: removed

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
desagar: I think this was simply an artifact of copying the JenkinsService as a starting point. We didn't think about when to use it or not.
Looking at the JenkinsService now, I see that a number of the calls use Hystrix but a handful do not - these seem to be the ones that retrieve data on a single item (e.g. get build by build number), and the ones that do POSTs (start/stop a build etc)

Would your suggestion be to follow a similar pattern and use Hystrix for most of the http calls?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitor.groovy
```diff
@@ -0,0 +1,335 @@
...
+                } )
+            }
+        } catch (e) {
+            log.error("Error processing runs for Wercker[{}]", kv("master", master), e)
+        }
+        log.debug("Took ${System.currentTimeMillis() - startTime}ms to retrieve Wercker pipelines (master: {})", kv("master", master))
+        return new PipelinePollingDelta(master: master, items: delta)
+    }
+
+    Set<String> getConfiguredJobs(String master) {
```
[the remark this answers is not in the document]
shihchang: I guess the "pre-filtering" you meant is to configure some filter in the config (igor.yml). My understanding is that the config change is done through halyard and may need to restart the entire spinnaker; restarting spinnaker would be an usability issue for production deployment.

To do the pre-filtering, we would need the list of ci triggers and stages that has been enabled in all the spinnaker pipelines. I am looking at the BuildEventMonitor but cannot find a simple way to get all ci triggers and stages. Would you clarify what in the echo that I should use and you would be more comfortable with? What I can tell is to let PipelineCache subscribes front50, polls and store the triggers and stages in the PipelineCache so that igor can query it through some new API, but that would be a new API and new logic to build, and I would consider it more complicate than my current approach. Another possible way is to get the pipelineConfigs from gate.controllers.PipelineConfigController.getAllPipelineConfigs; is this better for you?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitor.groovy
```diff
@@ -0,0 +1,335 @@
...
+        Set<String> jobs = []
+        front50.get().getAllPipelineConfigs().forEach({ pipeline ->
+            pipeline['triggers'].forEach({ trigger ->
+                if (trigger['enabled'] && trigger['type'] == 'wercker' && trigger['master'] == master) {
+                    log.debug "configured Trigger for ${master} pipeline: ${pipeline['application']} ${pipeline['name']} ${trigger}"
+                    jobs.add(trigger['job'])
+                }
+            })
+            pipeline['stages'].forEach({ stage ->
```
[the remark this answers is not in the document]
shihchang: I thought the stage is also using this polling in order to know status of the ci task.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitor.groovy
shihchang: Fixed. The comparators are now constants (of Run) to be reused.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitor.groovy
shihchang: Fixed elvis

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitor.groovy
shihchang: Fixed. Originally when echoService is NOT present, cache.setEventPosted may still get called; this may still be an issue in Jenkins as we used Jenkins as example.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerCache.java
shihchang: Fixed. The comparators are now constants (of Run) to be reused.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerCache.java
shihchang: Good point. Does spinnaker/halyard allows two ci masters of different types to use the same name? Anyway I added "_wercker" to prefix to avoid conflict.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/model/Application.groovy
shihchang: fixed

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/model/Application.groovy
shihchang: It would still need to import groovy.transform.ToString, right?
Application.java:10: error: cannot find symbol
@ToString() public class Application
 ^
  symbol: class ToString

Fixed.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/model/Run.groovy
shihchang: removed.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/model/Pipeline.groovy
shihchang: fixed.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/model/Run.groovy
shihchang: fixed.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/build/InfoControllerSpec.groovy
shihchang: I saw an issue in InfoController that the line of buildMasters.filteredMap(BuildServiceProvider.JENKINS)[master] could cause NPE when
buildMasters.filteredMap(BuildServiceProvider.JENKINS) returns null, instead of empty map, which I guess is for the case that there is no jenkins master. I am not sure if this null would happen only in the Mock(BuildMasters) or also in real BuildMasters. Those two lines of adding Jenkins and Wercker were to avoid the NPE. I changed 
InfoController to avoid NPE in case the real BuildMasters.filteredMap does return null.

I reverted this test to only adding the Jenkins though that line of adding jenkins is no longer needed (to avoid NPE). I also added a test for wercker.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSpec.groovy
shihchang: removed.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSpec.groovy
shihchang: fixed.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSpec.groovy
shihchang: Moved the duplicated setup to mockBuildMasters().

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSpec.groovy
shihchang: Moved the duplicated setup to mockBuildMasters().

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSpec.groovy
shihchang: Added the assertion that the eventPosted is from lastest finished one.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSpec.groovy
shihchang: fixed

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSchedulingSpec.groovy
shihchang: removed

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSchedulingSpec.groovy
shihchang: removed unused echoService.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSchedulingSpec.groovy
shihchang: fixed.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitorSchedulingSpec.groovy
shihchang: fixed.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitor.groovy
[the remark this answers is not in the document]
shihchang: Corrected - Wercker uses buildID (runId) but does not have sequential buildNumber for each build(run). So we are using the polling to build a map of string runID to sequential buildNumber in the cache which the stage will need to use it to get the status of the build/run.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerBuildMonitor.groovy
```diff
@@ -0,0 +1,335 @@
...
+            log.error("Error processing runs for Wercker[{}]", kv("master", master), e)
+        }
+        log.debug("Took ${System.currentTimeMillis() - startTime}ms to retrieve Wercker pipelines (master: {})", kv("master", master))
+        return new PipelinePollingDelta(master: master, items: delta)
+    }
+
+    Set<String> getConfiguredJobs(String master) {
+        Set<String> jobs = []
+        front50.get().getAllPipelineConfigs().forEach({ pipeline ->
+            pipeline['triggers'].forEach({ trigger ->
```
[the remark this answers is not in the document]
shihchang: Fixed, they are all in the format of obj.key now.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerCache.java
ezimanyi: I'm actually not sure if Halyard would let you configure that, but adding the "_wercker" prefix as you've done is safer anyway.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
ezimanyi: In general, I think using Hystrix is a good idea for communicating with systems outside of Spinnaker's control (or even outside of the control of the particular microservice in question), as it can help avoid an outage in the external system spilling into Spinnaker, and allows graceful degradation (by supplying fallback functions).  So I'd probably recommend using it.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/build/InfoControllerSpec.groovy
ezimanyi: Ok, that makes sense---both the fix for the NPE in `InfoController.groovy` and the updates to the tests; thanks for the explanation.

ezimanyi:
I've finished reading through the code, as well as the changes in the last few days---other than the front50 dependency, which I'd still like to discuss, everything looks good so this should be ready to go once we resolve the front50 issue.

My concern is less the actual dependency of igor -> front50, and more a question of how Igor now has knowledge of how pipelines are constructed so this is breaking an abstraction layer that we had before.  (Prior to this, Igor had no knowledge of what a "stage" was, or what a "trigger" was, it just knew it had to poll a CI system and send a message to Echo with the builds.)  If at all possible, I'd like to keep Igor unaware of these details, and instead use an API provided by another service to ask "what are the Wercker jobs I should poll?".  Or more generally just something like `getConfiguredCIJobs` that would return a structure like:
```
{
  jenkins: {
     master1: ['job1', 'job2'],
     master2: ['job1', 'job2'],
  },
  wercker: { 
     wercker_master1: ['job1', 'job2'],
     wercker_master2: ['job1', 'job2'],
  },
  ...
}
```
so that Igor could pass relevant jobs to Wercker, without needing to now how to derive these from the pipeline config.

I could see such an API being either in Front50 or Echo; on the one hand, Front50 is the authoritative source for pipelines so might make sense there.  On the other hand, Echo (in particular `echo-pipelinetriggers`) is the authority on when a particular build should trigger a pipeline...so it might be better suited to report on what CI jobs are relevant.  (This is somewhat similar to the existing `getSubscriptions` in echo where we ask it what pubsub subscriptions it knows about.) . I'm probably leaning a bit more towards echo at this point.

Finally, regarding the code where Igor is also looking for Wercker stages (https://github.com/spinnaker/igor/pull/297/files#diff-dfae39435ac8ec0f9ac5e0709c6b04a1R159), I see two options:
* This code could also call an API to ask what builds are configured as Wercker stages
* We could instead just go directly to Wercker rather than the cache when polling for completion of a Wercker stage (which is what Jenkins does https://github.com/spinnaker/igor/blob/7dbdaaf384805bdb1b10ab907c7178edd94b678d/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/service/JenkinsService.groovy#L99); given that at that point we know the specific build we're interested in, it should be an inexpensive all to ask Wercker about that job.

I'd probably suggest going with the second option, as I think the performance benefit of hitting the cache will probably not be worth the additional complexity of tracking which stages use Wercker jobs.

Let me know your thoughts on the above, and we can figure out the best approach.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/wercker/WerckerService.groovy
shihchang: all using Hystrix now.

PR #302, opened 2018-09-05
chore(web): Cleanup rest exception handlers

robzienert:
Just a bunch of exception handler cleanup.

There's a whole lot of referencing exception classes defined in controllers from lower-level classes, which I'd like to cleanup in a separate thing.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -37,7 +37,7 @@ allprojects {
     sourceSets.main.groovy.srcDirs += ["src/main/java"]
 
     ext {
-        spinnakerDependenciesVersion = project.hasProperty('spinnakerDependenciesVersion') ? project.property('spinnakerDependenciesVersion') : '0.161.5'
+        spinnakerDependenciesVersion = project.hasProperty('spinnakerDependenciesVersion') ? project.property('spinnakerDependenciesVersion') : '1.0.12'
     }
 
     def checkLocalVersions = [spinnakerDependenciesVersion: spinnakerDependenciesVersion]
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
@@ -19,21 +19,24 @@ package com.netflix.spinnaker.igor.build
 
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spinnaker.igor.build.model.GenericBuild
+
 import com.netflix.spinnaker.igor.jenkins.client.model.JobConfig
 import com.netflix.spinnaker.igor.jenkins.service.JenkinsService
 import com.netflix.spinnaker.igor.model.BuildServiceProvider
 import com.netflix.spinnaker.igor.service.ArtifactDecorator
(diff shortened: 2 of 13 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/auth/AuthRequestInterceptor.groovy
```diff
@@ -65,7 +65,6 @@ class AuthRequestInterceptor implements RequestInterceptor {
         }
 
         String toString() {
-            log.debug("Including Basic digest in Authorization header")
```
robzienert: This is spewing a ton of log messages into igor for every request.

build.gradle
```diff
@@ -37,7 +37,7 @@ allprojects {
     sourceSets.main.groovy.srcDirs += ["src/main/java"]
 
     ext {
-        spinnakerDependenciesVersion = project.hasProperty('spinnakerDependenciesVersion') ? project.property('spinnakerDependenciesVersion') : '0.161.5'
+        spinnakerDependenciesVersion = project.hasProperty('spinnakerDependenciesVersion') ? project.property('spinnakerDependenciesVersion') : '1.0.8'
```
ajordens: Might as well go all the way current? `1.0.12` I think?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
```diff
@@ -82,7 +84,7 @@ class BuildController {
 
             return build
         } else {
-            throw new MasterNotFoundException("Master '${master}' not found")
+            throw new NotFoundException("Master '${master}' not found")
```
ajordens: 👍

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/auth/AuthRequestInterceptor.groovy
ajordens: 👍

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/bitbucket/CommitController.groovy
```diff
@@ -68,11 +70,11 @@ class CommitController extends AbstractCommitController {
       }
     } catch (RetrofitError e) {
       if (e.getKind() == RetrofitError.Kind.NETWORK) {
-        throw new RuntimeException("Could not find the server ${bitBucketMaster.baseUrl}")
+        throw new NotFoundException("Could not find the server ${bitBucketMaster.baseUrl}")
```
ajordens: Is there a common cause for the `NETWORK` error? I wouldn't necessarily equate it to a 404 unless is a blatant misconfiguration or something and the config is pointed at an endpoint that doesn't respond?

robzienert: Yeah, I'm not really sure what's going on in this controller, it's always had this logic since the class was added. I'm going to go out on a limb and just delete that conditional.

PR #306, opened 2018-10-03
chore(*): Reduce logging noise; convert some logs to metrics

robzienert:
Cleaning up some more logs in igor. Moved some things to metrics that should be metrics, changed log levels to trace for some trace-worthy level logs and removed some others.


ajordens:
👍 to noise reduction

PR #321, opened 2018-10-17
feat(core): Apply artifact filtering in Igor

ezimanyi:
BREAKING CHANGE: This is a breaking change for users who have manually set BuildArtifactFilter.maxArtifacts or BuildArtifactFilter.preferredArtifacts as environment variables to Orca. As the filtering logic is being moved to Igor, these environment variables need to be set on the Igor process.

All current calls to get the build info for a particular build post-process by filtering the artifacts with BuildArtifactFilter.  Rather than require all callers to post-process, move this logic from Orca to Igor so that the API returns already-filtered artifacts.


ezimanyi:
This change just takes the `BuildArtifactFilter` that was in Orca and moves it to Igor so it can be applied before returning the details of a build.  I'm going to be adding some calls to Echo to retrieve build status, and needed the same filtering logic; given that all calls to get the build details do this filtering it seemed to make the most sense to push this into Igor rather than duplicate the filtering.  (It's also something that seems like it would fall within Igor's purview, given that it handles looking at CI build artifacts.)

If this looks good, then I'll remove these corresponding modules from Orca.

robzienert:
I'm into it, although if it's too big of a lift, it would be nice to simultaneously convert to Java.

ezimanyi:
@robzienert: Good idea; push a commit to convert to Java.

@gardleopard : Do you know if anyone at Schibsted is using this configuration parameter?

PR #331, opened 2018-11-16
fix(travis): grace period before triggering with new travis builds

gardleopard:
The travis api needs some time after a build is finished in order to have
the build data present in the apis.

This is configurable through `travis.newBuildGracePeriod`


gardleopard:
@spinnaker/oss-approvers PTAL

jkschneider:
LGTM

PR #333, opened 2018-11-21
fix(jenkins): replace SimpleXML with Jackson for XML deserialization

dreynaud:
I am now suspecting that SimpleXMLConverter is responsible for the leaked
connections. I haven't witnessed any leaked connections anymore after
switching to Jackson.


igor-web/igor-web.gradle
```diff
@@ -33,6 +33,7 @@ dependencies {
 
     compile 'org.yaml:snakeyaml:1.15'
     compile 'com.squareup.retrofit:converter-simplexml:1.9.0'
```
dreynaud: I would love to drop that dependency, as it is no longer used in the Jenkins model classes. However annotations from simplexml are still used in the Travis model classes, which I find a bit surprising since the Travis API seems JSON based and already using Jackson. @gardleopard do you have more context on this? I suspect these annotations are actually unused but I don't have a good way to test that theory

igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/client/model/ParameterDefinition.groovy
```diff
@@ -0,0 +1,61 @@
...
+    public String getDefaultValue() {
+        return defaultParameterValue.getValue()
+    }
+
+    // used when serializing
+    @XmlElement(name = "defaultName")
+    public String getDefaultName() {
+        return defaultParameterValue.getName()
+    }
+}
```
robzienert: Shoulda java'd this

PR #341, opened 2018-12-12
fix(http): make the getProperties endpoint return HTTP 404 status

robfletcher:
I suspect this is only the tip of the iceberg but trying to get relevant status codes / messages out of endpoints when they stray from the happy path seems worthwhile.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
@@ -17,9 +17,10 @@
 
 package com.netflix.spinnaker.igor.build
 
+import java.util.concurrent.ExecutorService
+import javax.servlet.http.HttpServletRequest
 import com.fasterxml.jackson.databind.ObjectMapper
 import com.netflix.spinnaker.igor.build.model.GenericBuild
-
 import com.netflix.spinnaker.igor.jenkins.client.model.JobConfig
 import com.netflix.spinnaker.igor.jenkins.service.JenkinsService
 import com.netflix.spinnaker.igor.model.BuildServiceProvider
@@ -32,21 +33,13 @@ import groovy.transform.InheritConstructors
 import groovy.util.logging.Slf4j
 import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.http.HttpStatus
-import org.springframework.web.bind.annotation.PathVariable
-import org.springframework.web.bind.annotation.RequestMapping
-import org.springframework.web.bind.annotation.RequestMethod
-import org.springframework.web.bind.annotation.RequestParam
-import org.springframework.web.bind.annotation.ResponseStatus
-import org.springframework.web.bind.annotation.RestController
+import org.springframework.web.bind.annotation.*
 import org.springframework.web.servlet.HandlerMapping
 import org.yaml.snakeyaml.Yaml
-import org.yaml.snakeyaml.constructor.SafeConstructor;
+import org.yaml.snakeyaml.constructor.SafeConstructor
 import retrofit.RetrofitError
-
-import javax.servlet.http.HttpServletRequest
-import java.util.concurrent.ExecutorService
-
 import static net.logstash.logback.argument.StructuredArguments.kv
+import static org.springframework.http.HttpStatus.NOT_FOUND
 
 @Slf4j
 @RestController
@@ -71,7 +64,8 @@ class BuildController {
     ArtifactDecorator artifactDecorator
 
     @RequestMapping(value = '/builds/status/{buildNumber}/{master:.+}/**')
-    GenericBuild getJobStatus(@PathVariable String master, @PathVariable Integer buildNumber, HttpServletRequest request) {
+    GenericBuild getJobStatus(@PathVariable String master, @PathVariable
+        Integer buildNumber, HttpServletRequest request) {
         def job = (String) request.getAttribute(
             HandlerMapping.PATH_WITHIN_HANDLER_MAPPING_ATTRIBUTE).split('/').drop(5).join('/')
         if (buildMasters.map.containsKey(master)) {
@@ -97,13 +91,13 @@ class BuildController {
     }
 
     @RequestMapping(value = '/builds/queue/{master}/{item}')
-    Object getQueueLocation(@PathVariable String master, @PathVariable int item){
+    Object getQueueLocation(@PathVariable String master, @PathVariable int item) {
         if (buildMasters.filteredMap(BuildServiceProvider.JENKINS).containsKey(master)) {
 
             try {
                 return buildMasters.map[master].getQueuedItem(item)
             } catch (RetrofitError e) {
-                if (e.response?.status == HttpStatus.NOT_FOUND.value()) {
+                if (e.response?.status == NOT_FOUND.value()) {
                     throw new NotFoundException("Queued job '${item}' not found for master '${master}'.")
                 }
                 throw e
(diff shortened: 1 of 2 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
```diff
@@ -226,20 +220,15 @@ class BuildController {
...
             try {
                 def jenkinsService = buildMasters.map[master]
-                String path = getArtifactPathFromBuild(jenkinsService, job, buildNumber, fileName)
-
-                if (path == null) {
-                    log.error("Unable to get igorProperties: Could not find build artifact matching requested filename '{}' on '{}' build '{}",
-                            kv("fileName", fileName), kv("master", master), kv("buildNumber", buildNumber))
-                    return map
-                }
+                String path = getArtifactPathFromBuild(jenkinsService, master, job, buildNumber, fileName)
```
robfletcher: Was previously just returning an empty map if the artifact path is not found.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
```diff
@@ -256,6 +245,8 @@ class BuildController {
                 } finally {
                     propertyStream.close()
                 }
+            } catch (NotFoundException e) {
+                throw e
```
robfletcher: Really don't want to catch & ignore this type of error

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
```diff
@@ -271,24 +262,27 @@ class BuildController {
...
-                    it.fileName == fileName
-                }
-                if (artifact) {
-                    return artifact.relativePath
-                }
-                log.info("Could not find artifact {} for job {} on build #{}", fileName, job, buildNumber)
-                throw new ArtifactNotFoundException()
-            }, 5, 2000, false)
-        } catch (ArtifactNotFoundException ignored) {
-            return null
```
robfletcher: I removed this empty catch block so the exception propagates out of here

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
```diff
@@ -271,24 +262,27 @@ class BuildController {
...
+                log.error("Unable to get igorProperties: Could not find build artifact matching requested filename '{}' on '{}' build '{}",
+                    kv("fileName", fileName), kv("master", master), kv("buildNumber", buildNumber))
+                throw new ArtifactNotFoundException(master, job, buildNumber, fileName)
+            }
+        }, 5, 2000, false)
     }
 
-    private static class ArtifactNotFoundException extends RuntimeException {}
+    @ResponseStatus(NOT_FOUND)
+    private static class ArtifactNotFoundException extends NotFoundException {
```
robfletcher: Extending `NotFoundException` means we get Kork's default error handling & thus HTTP 404.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/build/BuildControllerSpec.groovy
```diff
@@ -106,7 +101,7 @@ class BuildControllerSpec extends Specification {
             .accept(MediaType.APPLICATION_JSON)).andReturn().response
 
         then:
-        3 * buildMasters.map >> [MASTER: service]
+        3 * buildMasters.map >> [(MASTER): service]
```
robfletcher: I made a bunch of changes like this as it appears it was only working by accident before because the variable name `MASTER` is the same as its value. If you don't use braces here Groovy treats the key portion as a string literal _not_ a variable reference.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/build/BuildControllerSpec.groovy
```diff
@@ -175,25 +170,40 @@ class BuildControllerSpec extends Specification {
...
+        buildMasters.filteredMap(BuildServiceProvider.TRAVIS) >> [:]
+        buildMasters.map >> [(MASTER): jenkinsService]
+
+        expect:
+        mockMvc.perform(
+            get("/builds/properties/${BUILD_NUMBER}/${FILE_NAME}/badMaster/${JOB_NAME}")
+                .accept(MediaType.APPLICATION_JSON))
+            .andExpect(status().isNotFound())
+            .andReturn().response
+    }
```
robfletcher: Thought it was worth checking this condition also results in a 404.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/build/BuildControllerSpec.groovy
```diff
@@ -175,25 +170,40 @@ class BuildControllerSpec extends Specification {
...
+
     void 'get properties of a build with a bad filename'() {
         given:
-        5 * jenkinsService.getBuild(JOB_NAME, BUILD_NUMBER) >> new Build(
+        jenkinsService.getBuild(JOB_NAME, BUILD_NUMBER) >> new Build(
```
robfletcher: This is a stub not a mock, we don't care

igor-web/src/test/groovy/com/netflix/spinnaker/igor/build/BuildControllerSpec.groovy
```diff
@@ -175,25 +170,40 @@ class BuildControllerSpec extends Specification {
...
+        mockMvc.perform(
             get("/builds/properties/${BUILD_NUMBER}/${FILE_NAME}/${MASTER}/${JOB_NAME}")
-            .accept(MediaType.APPLICATION_JSON)).andReturn().response
-
-        then:
-        1 * buildMasters.filteredMap(BuildServiceProvider.JENKINS) >> [MASTER: jenkinsService]
-        1 * buildMasters.map >> [MASTER: jenkinsService]
-        response.contentAsString == "{}"
+                .accept(MediaType.APPLICATION_JSON))
+            .andExpect(status().isNotFound())
```
robfletcher: Actually check the status (it was HTTP 200 before)

robfletcher:
@dreynaud looks like you touched this most recently.

I had a user today who specified the properties file name as `build/spinnaker.properties`. We don't want the path but Igor does not give us an HTTP 404 back so it took me a while to track down.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
TIME-GATE: hi, when I delpoy spinnaker on k8s, the igor component show error:

Caused by: retrofit.converter.ConversionException: org.simpleframework.xml.core.ValueRequiredException: Unable to satisfy @org.simpleframework.xml.ElementList(entry=, data=false, inline=true, name=job, type=void, required=true, empty=true) on field 'list' private java.util.List com.netflix.spinnaker.igor.jenkins.client.model.ProjectsList.list for class com.netflix.spinnaker.igor.jenkins.client.model.ProjectsList

can you give me some clues, I have try to solve this problem by build docker image from source code, but failed, thanks!

robfletcher: Can you raise this in a relevant forum? Either the #dev or #kubernetes channels on the Spinnaker Slack would seem appropriate. This issue and this repository have absolutely nothing to do with Spinnaker's support for Kubernetes or any problems you might be experiencing with that.

PR #344, opened 2018-12-21
 fix(docker): Update cache when a null tag becomes non-null

ezimanyi:
We currently don't update the docker image cache when either the prior or the current digest is null. This means that any tag that was present before trackDigests was enabled will always have its digest set to null. Change this so we do set the digest any time the new one is not null.

In order to avoid triggering multiple pipelines when turning trackDigests on, we won't send an event to echo when updating from a null to non-null digest (as we don't actually know if it was updated), we'll only update the cache.


ezimanyi:
Addresses the issue here: https://community.spinnaker.io/t/trigger-on-push-to-docker-registry/916/10

igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerMonitor.groovy
```diff
@@ -202,5 +207,28 @@ class DockerMonitor extends CommonPollingMonitor<ImageDelta, DockerPollingDelta>
     private static class ImageDelta implements DeltaItem {
         String imageId
         TaggedImage image
+        boolean sendEvent = true
+    }
+
+    private static class UpdateType {
```
lwander: Nice

PR #357, opened 2019-01-26
fix(jenkins): Add Spinnaker user-agent string to Jenkins requests

marchello2000:
Add `Spinnaker-igor` for User-Agent header to all requests to Jenkins
following the convention in clouddriver.


marchello2000:
> Got some tests failing

Doesn't fail locally no matter what I try, rerunning the build "fixed" it

PR #366, opened 2019-02-04
feat(jenkins/queue): add ready flag to queue response

gardleopard:
Currently one has to check for a null to know if a build number is set
or not. Setting a ready Boolean will make it more clear that jenkins is
spending some time setting a build number for a queued item.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/client/model/QueuedJob.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/client/model/QueuedJob.groovy
@@ -29,6 +29,11 @@ class QueuedJob {
     Integer getNumber() {
         return executable?.number
     }
+
+    @XmlElement(name = 'ready')
+    boolean getReady() {
+        executable?.number ? true : false
+    }
 }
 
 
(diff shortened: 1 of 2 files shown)


gardleopard:
@ajordens it adds it to the queue, but it takes a little while before build number is populated. This is currently handled in orca https://github.com/spinnaker/orca/blob/7c4896117fdcd6930cadaa82bfdf5dcb8383f89b/orca-igor/src/main/groovy/com/netflix/spinnaker/orca/igor/tasks/MonitorQueuedJenkinsJobTask.groovy#L48

I did not like the ready name either, however queued can also be misleading here right?

ajordens:
True ... is the goal just to simplify that check in `orca` or is there elsewhere that we're thrown by the fact that we need to wait for the number to be populated?

I don't know that adding a new attribute is much better than doing the null check, it's context that you still have to be aware of.

🤷‍♂️ No strong opinion against it being merged, just unsure it'll make things any more clearer than a comment at the call site in orca.

gardleopard:
I faced this when using igor without using spinnaker. I had to build logic that I thought igor handled for me. That's where I got the inspiration to add this field.

ajordens:
Oops, sorry ... this slipped off my radar.

To be honest I'm not sure how much value this adds ... you still need to know that `ready` would need to be checked prior to `number`.

Not convinced we need to merge, but if you feel it's a value add I don't object that strongly.

Feel free to merge or close.

gardleopard:
I'm looking into how I could do this in a better way.

PR #374, opened 2019-02-21
feat(buildmasters): Permission support for build masters (CI's)

jervi:
New feature to let build masters be restricted to people with certain roles.

Read permission is required to see the build master at all, and thus to use triggers.
Write permission is only required to be able to trigger a build with the build master.

Adds a permissions field to the config in Igor:
```
travis:
  repositorySyncEnabled: true
  masters:
  - name: travis
    address: https://travis.com/api
    baseUrl: https://travis.com
    githubToken: xxxxxxx
    numberOfRepositories: 100
    permissions:
      READ:
      - foo
      - bar
      WRITE:
      - bar
```

In the example above, users with the `foo` or `bar` roles will be able to see the build master and use it as a trigger, users with the `bar` role will additionally be able to trigger builds. Users without these roles will not see the build master in Deck at all.

This feature is useful for big Spinnaker installations where companies and/or teams want to bring their own CI's into Spinnaker.

https://github.com/spinnaker/echo/pull/467, https://github.com/spinnaker/orca/pull/2673 and https://github.com/spinnaker/fiat/pull/355 is also a part of this feature and depends on this PR. Halyard support in https://github.com/spinnaker/halyard/pull/1224.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/InfoController.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/InfoController.groovy
@@ -24,13 +24,19 @@ import com.netflix.spinnaker.igor.config.TravisProperties
 import com.netflix.spinnaker.igor.config.WerckerProperties
 import com.netflix.spinnaker.igor.jenkins.service.JenkinsService
 import com.netflix.spinnaker.igor.model.BuildServiceProvider
+import com.netflix.spinnaker.igor.service.BuildService
 import com.netflix.spinnaker.igor.service.BuildServices
 import com.netflix.spinnaker.igor.wercker.WerckerService
 import com.netflix.spinnaker.kork.web.exceptions.NotFoundException
 import groovy.util.logging.Slf4j
-import org.springframework.beans.factory.annotation.Autowire
 import org.springframework.beans.factory.annotation.Autowired
-import org.springframework.web.bind.annotation.*
+import org.springframework.security.access.prepost.PostFilter
+import org.springframework.security.access.prepost.PreAuthorize
+import org.springframework.web.bind.annotation.PathVariable
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestMethod
+import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.bind.annotation.RestController
 import org.springframework.web.servlet.HandlerMapping
 
 import javax.servlet.http.HttpServletRequest
@@ -64,45 +70,25 @@ class InfoController {
     ConcourseProperties concourseProperties
 
     @RequestMapping(value = '/masters', method = RequestMethod.GET)
-    List<Object> listMasters(
-        @RequestParam(value = "showUrl", defaultValue = "false") String showUrl,
-        @RequestParam(value = "type", defaultValue = "") String type) {
+    @PostFilter("hasPermission(filterObject, 'BUILD_SERVICE', 'READ')")
+    List<String> listMasters(@RequestParam(value = "type", defaultValue = "") String type) {
 
-        BuildServiceProvider providerType = (type == "") ? null :
-            BuildServiceProvider.valueOf(type.toUpperCase())
-
-        if (showUrl == 'true') {
-            List<Object> masterList = []
-            addMaster(masterList, providerType, jenkinsProperties,  BuildServiceProvider.JENKINS)
-            addMaster(masterList, providerType, travisProperties,   BuildServiceProvider.TRAVIS)
-            addMaster(masterList, providerType, gitlabCiProperties, BuildServiceProvider.GITLAB_CI)
-            addMaster(masterList, providerType, werckerProperties,  BuildServiceProvider.WERCKER)
-            addMaster(masterList, providerType, concourseProperties, BuildServiceProvider.CONCOURSE)
-            return masterList
+        BuildServiceProvider providerType = (type == "") ? null : BuildServiceProvider.valueOf(type.toUpperCase())
+        //Filter by provider type if it is specified
+        if (providerType) {
+            return buildServices.getServiceNames(providerType)
         } else {
-            //Filter by provider type if it is specified
-            if (providerType) {
-                return buildServices.getServiceNames(providerType)
-            } else {
-                return buildServices.getServiceNames()
-            }
+            return buildServices.getServiceNames()
         }
     }
 
-    void addMaster(masterList, providerType, properties, expctedType) {
-        if (!providerType || providerType == expctedType) {
-            masterList.addAll(
-                properties?.masters.collect {
-                    [
-                        "name"         : it.name,
-                        "address"      : it.address
-                    ]
-                }
-            )
-        }
+    @RequestMapping(value = '/buildServices', method = RequestMethod.GET)
+    List<BuildService> getAllBuildServices() {
+        buildServices.allBuildServices
     }
 
     @RequestMapping(value = '/jobs/{master:.+}', method = RequestMethod.GET)
+    @PreAuthorize("hasPermission(#master, 'BUILD_SERVICE', 'READ')")
     List<String> getJobs(@PathVariable String master) {
         def buildService = buildServices.getService(master)
         if (buildService == null) {
@@ -114,7 +100,7 @@ class InfoController {
             def jobList = []
             def recursiveGetJobs
 
-            recursiveGetJobs = { list, prefix="" ->
+            recursiveGetJobs = { list, prefix = "" ->
                 if (prefix) {
                     prefix = prefix + "/job/"
                 }
@@ -138,6 +124,7 @@ class InfoController {
     }
 
     @RequestMapping(value = '/jobs/{master:.+}/**')
+    @PreAuthorize("hasPermission(#master, 'BUILD_SERVICE', 'READ')")
     Object getJobConfig(@PathVariable String master, HttpServletRequest request) {
         def job = (String) request.getAttribute(
             HandlerMapping.PATH_WITHIN_HANDLER_MAPPING_ATTRIBUTE).split('/').drop(3).join('/')
@@ -147,4 +134,5 @@ class InfoController {
         }
         return service.getJobConfig(job)
     }
+
 }
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/service/BuildOperations.java
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/service/BuildOperations.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2019 Schibsted ASA.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ *
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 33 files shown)


jervi:
@spinnaker/reviewers PTAL

jervi:
I've done some major refactorings after discussing this with @dibyom. All authentication checks have been moved from Gate to Igor, and it's now using the Fiat api properly (I've added a new resource type to Fiat and Igor is using standard Spring annotations to check authentication).

igor-web/src/main/groovy/com/netflix/spinnaker/igor/filters/CorsFilter.groovy
```diff
@@ -1,56 +0,0 @@
-/*
```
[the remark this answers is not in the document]
jervi: Yes. It comes with Fiat, so Spring complains about it being loaded twice if it is present here as well

dibyom:
Permission checks looks fine but I'm not super familiar with Igor so tagging @ezimanyi who has done looked at it recently

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/InfoController.groovy
```diff
@@ -59,44 +65,27 @@ class InfoController {
...
-            masterList.addAll(
-                properties?.masters.collect {
-                    [
-                        "name"         : it.name,
-                        "address"      : it.address
-                    ]
-                }
-            )
+    @RequestMapping(value = '/buildServices', method = RequestMethod.GET)
+    List<BuildServiceView> getAllBuildServices() {
```
ezimanyi: Can the logic to construct this be moved to `BuildServices.java`?  I just refactored a lot of logic out of the controllers, trying to push them into Java classes, and would like to avoid moving more logic back to the controllers.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/filters/CorsFilter.groovy
ezimanyi: I know this was discussed above, but are you sure that the bean that Spring is actually correctly configured?  I'm pretty sure that just having the Spring-provided CorsFilter bean exist will not be enough to set the CORS headers without adding `@CrossOrigin` annotations, and even then we'll need to be very careful that the behavior stays the same as this custom class.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/service/BuildServices.java
```diff
@@ -46,4 +47,8 @@ public BuildService getService(String name) {
             .sorted()
             .collect(Collectors.toList());
     }
+
+    public Map<String, BuildService> getAllBuildServices() {
```
ezimanyi: As noted above, I'd prefer if any logic/processing of build services could happen directly in this class, with the result returned to callers.  It looks like this could be replaced with a function that returns a `List<BuildServiceView>`.

Prior to a recent refactor I did, most callers were just grabbing the map and doing their own (often duplicated) logic so I'd like to avoid adding a public method here that just returns the map, in favor of having callers ask for the data they want.

igor-web/src/main/java/com/netflix/spinnaker/igor/config/BuildMasterProperties.java
```diff
@@ -0,0 +1,32 @@
+/*
```
ezimanyi: This is minor, but it seems like we're standardizing on `BuildService` and slowly migrating the jenkins-specific "build master" terminology to "build service".  Based on that it might make more sense to call this `BuildServiceProperties`.

ezimanyi:
Also, forgot to include this in the above comment, but I have some questions around how this will work in the context of pipeline triggers.  Many of the calls to Igor come from Echo in the context of preparing a trigger before sending it to Orca to execute.

I don't think that Echo will have any authorization at that point---so there will be no authorization context to pass along to Igor.  Eventually once Echo figures out what pipeline to run and sends it to Orca, Orca will apply the authorization context of the pipeline's service account---but none of that has happened in Echo yet.  So this change may end up causing any automated build triggers to fail as they won't have permission to call back to igor to get the build information.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/InfoController.groovy
```diff
@@ -59,44 +65,27 @@ class InfoController {
...
-            List<Object> masterList = []
-            addMaster(masterList, providerType, jenkinsProperties,  BuildServiceProvider.JENKINS)
-            addMaster(masterList, providerType, travisProperties,   BuildServiceProvider.TRAVIS)
-            addMaster(masterList, providerType, gitlabCiProperties, BuildServiceProvider.GITLAB_CI)
-            addMaster(masterList, providerType, werckerProperties,  BuildServiceProvider.WERCKER)
-            return masterList
+    @PostFilter("hasPermission(filterObject, 'BUILD_SERVICE', 'READ')")
+    List<String> listMasters(@RequestParam(value = "type", defaultValue = "") String type) {
+
+        BuildServiceProvider providerType = (type == "") ? null : BuildServiceProvider.valueOf(type.toUpperCase())
```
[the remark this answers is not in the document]
jervi: There's not a single occurrence of `showUrl` in the entire Spinnaker code base, and the Gate API does not forward this parameter, so I think it is safe to remove it.

igor-web/src/main/java/com/netflix/spinnaker/igor/config/BuildMasterProperties.java
jervi: Fixed

igor-web/src/main/groovy/com/netflix/spinnaker/igor/service/BuildServices.java
jervi: Fixed in 4bc8b65ea8083dfe7922eab43fd67dbd6f20f7fd

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/InfoController.groovy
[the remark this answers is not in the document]
ezimanyi: Sounds good---makes sense then!

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/InfoController.groovy
```diff
@@ -141,4 +131,16 @@ class InfoController {
         }
         return service.getJobConfig(job)
     }
+
+    class BuildServiceView {
```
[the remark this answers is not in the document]
jervi: Fixed

igor-web/src/main/groovy/com/netflix/spinnaker/igor/filters/CorsFilter.groovy
jervi: I'll double check...

jervi:
> Also, forgot to include this in the above comment, but I have some questions around how this will work in the context of pipeline triggers. Many of the calls to Igor come from Echo in the context of preparing a trigger before sending it to Orca to execute.
I've tested with manual triggers, Travis triggers and Cron triggers, and all have the necessary context. Echo seems to be aware of the `runAsUser` field, and propagates that to Igor. For manual users, the necessary headers are present, so it just works (with these changes present: 
https://github.com/spinnaker/echo/pull/467)

igor-web/src/main/groovy/com/netflix/spinnaker/igor/service/BuildServices.java
ezimanyi: Looks great, thanks!

ezimanyi:
I turned `HEADERS` level logging for retrofit in `echo`, and looked at the calls to `igor`.  When I trigger a pipeline manually, the authentication is propagated and I see the following headers in the calls to `igor`:
```
X-SPINNAKER-REQUEST-ID
X-SPINNAKER-USER-ORIGIN
X-SPINNAKER-USER
X-SPINNAKER-ACCOUNTS
```

If the job starts automatically because a Jenkins job completed, the calls to `igor` do not have either `X-SPINNAKER-USER` or `X-SPINNAKER-ACCOUNTS` set.  Since the initial call posting the event to `echo` had no authentication context, there is no authentication to propagate.  The code linked above to set `runUser` only sets it immediately before sending the pipieline to `orca`; this wouldn't be set while the trigger is being constructed (which is when many of these calls to `igor` happen).

I think we'll need to put some thought into exactly how authorization should work in the context of echo processing an event.  It might make sense for it to assume the role of the `runAsUser` on the trigger while it's making any external calls it needs to construct the trigger, but that's not something that's happening now as far as I can tell.

jervi:
Is this different for Travis? I placed thread breakpoints in all the controller methods in Igor, and every interaction had auth set, both for manual and automatic triggers

jervi:
@ezimanyi Echo -> Igor auth addressed in https://github.com/spinnaker/echo/pull/467/commits/a669dfdc602ebd064919554d2a5eb7ea17611c77

ezimanyi:
Looks good, the question around `CorsFilter` is the only thing outstanding from my perspective but would definitely still like someone from netflix to review as well.

jervi:
I've added back the CorsFilter class. I don't know why Igor initially failed when starting up, bot now it worked. I confirmed using the actuator /beans endpoint that the CorsFilter was loaded as a bean.
```json
      {
        "bean": "corsFilter",
        "aliases": [],
        "scope": "singleton",
        "type": "com.netflix.spinnaker.igor.filters.CorsFilter",
        "resource": "file [/Users/jorjervi/dev/spinnaker/igor/igor-web/out/production/classes/com/netflix/spinnaker/igor/filters/CorsFilter.class]",
        "dependencies": []
      }
```

jervi:
Btw, this version of Igor has a hard dependency on [this Fiat PR](https://github.com/spinnaker/fiat/pull/355) and that version of the fiat-api needs to be available in Igor. So don't merge before that PR is merged, tagged, updated in Spinnaker-dependencies, and Igor has gotten the new version of Spinnaker-dependencies.

dibyom:
The igor support in Fiat is behind a flag that is disabled by default. So it should be safe to merge that in. Though we do need a new fiat-api release for this to work

igor-web/src/main/groovy/com/netflix/spinnaker/igor/service/BuildOperations.java
```diff
@@ -0,0 +1,31 @@
...
+
+package com.netflix.spinnaker.igor.service;
+
+import com.netflix.spinnaker.igor.build.model.GenericBuild;
+import com.netflix.spinnaker.igor.build.model.GenericGitRevision;
+
+import java.util.List;
+import java.util.Map;
+
+public interface BuildOperations extends BuildService {
```
robzienert: I'd like to see class- & method-level docs for all interfaces introduced into the Spinnaker ecosystem.

igor-web/src/main/java/com/netflix/spinnaker/igor/config/BuildServerProperties.java
```diff
@@ -0,0 +1,32 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.igor.config;
+
+import com.netflix.spinnaker.fiat.model.resources.Permissions;
+
+import java.util.List;
+
+public interface BuildServerProperties<T extends BuildServerProperties.Host> {
```
robzienert: More docs.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/service/BuildOperations.java
jervi: Better?

PR #378, opened 2019-03-04
 refactor(jenkins): Move logic from BuildController to JenkinsService

ezimanyi:
There is a lot of Jenkins-specific logic in the `BuildController` module that is difficult to re-use and adds complexity to the controller.  Refactor this (and in the process convert to Java).


ezimanyi:
* test(ci): Remove implementation checks from tests

Many of the test on the CI classes test the implementation of
the classes rather than the interface---in particular by verifying
the number of calls to buildMasters. Replace these tests with an
actual BuildMasters populated with test data and remove checks on
how many times its methods are called.

* refactor(ci): Improve the contract around build services

The BuildMasters class currently has a public map, which is accessed
from everywhere. Give it an interface, addServices() to add services
and getService() to fetch a service, and made the implementation
details private.

* refactor(jenkins): Convert JenkinsService and BuildService to Java\

* refactor(jenkins): Move logic from BuildController to JenkinsService

The BuildController has a lot of logic that should be in the individual
build services. Move towards making this better by pushing some of the
Jenkins-specific logic into JenkinsService

ezimanyi:
I tried to do one thing in each of the commits, so it will probably be easier to read the commits separately.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
```diff
@@ -49,7 +52,7 @@ class BuildController {
     ExecutorService executor
 
     @Autowired
-    BuildMasters buildMasters
+    BuildServices buildMasters
```
jtk54: Maybe change the variable name to match the type?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/InfoController.groovy
```diff
@@ -49,7 +49,7 @@ class InfoController {
     JenkinsProperties jenkinsProperties
 
     @Autowired
-    BuildMasters buildMasters
+    BuildServices buildMasters
```
jtk54: Same here

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
```diff
@@ -94,31 +80,22 @@ class BuildController {
...
-                return buildService.getQueuedItem(item)
-            } catch (RetrofitError e) {
-                if (e.response?.status == NOT_FOUND.value()) {
-                    throw new NotFoundException("Queued job '${item}' not found for master '${master}'.")
-                }
-                throw e
-            }
-        } else if (buildService.buildServiceProvider() == BuildServiceProvider.TRAVIS) {
-            return buildService.queuedBuild(item)
+        if (buildService instanceof JenkinsService) {
```
jtk54: Maybe worth encapsulating the logic getting a JenkinsService into a `getJenkinsService()` since it's probably used for each API endpoint.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
```diff
@@ -94,31 +80,22 @@ class BuildController {
...
+            JenkinsService jenkinsService = (JenkinsService) buildService
+            return jenkinsService.queuedBuild(item);
+        } else if (buildService instanceof TravisService) {
+            TravisService travisService = (TravisService) buildService
+            return travisService.queuedBuild(item)
         }
+        return null
```
jtk54: Should we explicitly throw and exception here as opposed to returning null and letting the upstream (downstream?) service deal with it?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
ezimanyi: Pushed a new commit to rename this everywhere

jtk54: :+1:

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
```diff
@@ -17,55 +17,41 @@
...
-    RetrySupport retrySupport
+    private BuildServices buildMasters
 
     @Autowired(required = false)
-    BuildArtifactFilter buildArtifactFilter
+    private BuildArtifactFilter buildArtifactFilter
 
     @Autowired(required = false)
-    ArtifactDecorator artifactDecorator
+    private ArtifactDecorator artifactDecorator
```
emjburns: We tend to like classes to have the pattern of an autowired constructor instead of autowired fields, even though a lot of the older code has the autowired field pattern.

```
private BuildArtifactFilter buildArtifactFilter
private ArtifactDecorator artifactDecorator

@Autowired
public BuildController(BuildArtifactFilter buildArtifactFilter, ArtifactDecorator artifactDecorator) {
this.buildArtifactFilter = buildArtifactFilter
...
```
To make this work with `@Autowired(required = false)` you just wrap the constructor param in an optional, and then do any checking/error handling/default values in the constructor assignment.

If you're already doing major refactoring and turning classes into java this might be another little goodie to throw in. Not necessary though!

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
ezimanyi: Yes, that makes sense---will throw an exception here

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
ezimanyi: I'm not sure I fully understand.  Ideally we'd completely remove the switch statements in the controller by delegating more to the `BuildService` inheritance chain...but for new we're getting the service (regardless of the type) and just checking what type it is.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
ezimanyi: Good point, I sort of stopped here as this was getting to be a big PR, but I'm all for constructor injection so I'll add it in!

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
robzienert: Not sure I understand why we need to do any casting here? Is `queuedBuild(item)` not part of the `BuildService` interface? If not, should we introduce a `QueuingBuildService` that extends so we don't need to cast this around?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/exceptions/ArtifactNotFoundException.java
```diff
@@ -0,0 +1,29 @@
...
+
+import com.netflix.spinnaker.kork.web.exceptions.NotFoundException;
+import org.springframework.web.bind.annotation.ResponseStatus;
+
+import static org.springframework.http.HttpStatus.NOT_FOUND;
+
+@ResponseStatus(NOT_FOUND)
+public class ArtifactNotFoundException extends NotFoundException {
+    public ArtifactNotFoundException(String master, String job, Integer buildNumber, String fileName) {
+        super("Could not find build artifact matching requested filename '$fileName' on '$master/$job' build $buildNumber");
```
robzienert: I think you need `String.format` here.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
ezimanyi: It's not part of the interface---I did consider having `JenkinsBuildService` and `TravisBuildService` implement a new `QueuingBuildService` interface which I think is probably the right thing here---but tried to focus this PR more on adding actual typing to the controller via `instanceof` and leave removing the crazy if blocks to a future PR as this was getting to be a big PR.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/exceptions/ArtifactNotFoundException.java
ezimanyi: Ha, I do...conversion from groovy fail

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
robzienert: I can appreciate that. At least we're all on the same page as far as some next step improvements go. :)

PR #386, opened 2019-03-08
 feat(artifacts): Add igor endpoint to get artifacts from a build

ezimanyi:
refactor(artifacts): Move artifact logic from echo to igor

This commit just moves the logic for parsing artifacts from builds
from echo to igor.

feat(artifacts): Add igor endpoint to get artifacts from a build

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
@@ -17,6 +17,7 @@
 
 package com.netflix.spinnaker.igor.build
 
+import com.netflix.spinnaker.igor.artifacts.ArtifactExtractor
 import com.netflix.spinnaker.igor.build.model.GenericBuild
 import com.netflix.spinnaker.igor.exceptions.BuildJobError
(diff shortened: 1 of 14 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
```diff
@@ -74,8 +76,30 @@ class BuildController {
         if (buildArtifactFilter) {
             build.artifacts = buildArtifactFilter.filterArtifacts(build.artifacts)
         }
+        return build;
+    }
 
-        return build
+    @RequestMapping(value = '/builds/status/{buildNumber}/{master:.+}/**')
```
jtk54: What does the `**` correspond to here?

ezimanyi: The current pattern in the `BuildController` is to append the Jenkins job name (slashes and all) to the end of the URL, so this (and the `def job` line below) grab everything at the end and consider it to be the job.

I'm not sure why we don't encode the job name and consider it as a single query token (or even make it a parameter) but wanted to be consistent with the other endpoints.  It means you have to send the job name unencoded, which I learned [here](https://github.com/spinnaker/echo/pull/421).

PR #389, opened 2019-03-11
feat(core): Add support for an Artifactory Trigger

Jammy-Louie:
spinnaker/spinnaker#4069


igor-web/src/main/java/com/netflix/spinnaker/igor/artifactory/model/ArtifactoryArtifact.java
```diff
@@ -40,7 +40,7 @@ public Artifact toMatchableArtifact(ArtifactoryRepositoryType repoType) {
 
         return Artifact.builder().type("maven/file")
           .reference(group + ":" + artifactId + ":" + version)
-          .name(group + ":" + artifactId)
+          .name(group + ":" + artifactId + ":" + version)
```
jkschneider: I did intend to leave version off of this. Do we need it for any reason further upstream?

Jammy-Louie: done

PR #396, opened 2019-03-14
fix(web): make concourse wiring optional

asher:


asher:
This fixes an issue where igor fails to start if concourse isn't enabled.

PR #395, opened 2019-03-14
feat(core): add jenkins artifact type

Jammy-Louie:
spinnaker/spinnaker#4101

PR #398, opened 2019-03-15
Revert "feat(core): add jenkins artifact type (#395)"

emjburns:
@Jammy-Louie @jkschneider reverting this because it is causing lots of timeouts in our jenkins setup.

Please put this behind a feature flag that defaults to disabled. In the future, when you're touching core components of the services, please try and get another set of eyes from Netflix/Google/someone else running a large installation.


ezimanyi:
@spinnakerbot is not going to like this commit name, but I'm fine with it!

PR #399, opened 2019-03-15
fix(docker): fix gradle build step

bpowell:


bpowell:
related to https://github.com/spinnaker/clouddriver/pull/3475/files

PR #402, opened 2019-03-20
fix(concourse): Detect Concourse token expiry

jkschneider:
See https://github.com/concourse/concourse/issues/3558.

Concourse does not correctly produce a 401 when a token is invalid/expires when executing commands to retrieve teams and pipelines.

Calls to `/sky/userinfo` do return a 401 on invalid/expired token, so as a workaround we preface calls to resource listings with a call to `/sky/userinfo`.


jkschneider:

PR #404, opened 2019-03-21
fix(concourse): Tweak return type to support manual execution build listing

jkschneider:
Returning a `GenericBuild` now which properly supports listing builds available for manual execution:

image


jkschneider:
It's intentional. One needs to fetch builds with details, the other without. The `getBuilds` method that returns `Build` instead of `GenericBuild` is used by `ConcourseBuildMonitor` as well. Following the approach in the Travis implementation.

Ultimately when @robzienert rewrites Igor he'll have to separate these interfaces and the return types will get more clear.

jkschneider:

PR #407, opened 2019-03-22
Artifact services

emjburns:
(I changed igor's default tab size to 2 like all our other services. This will only affect new code and also auto-indenting).

Adding a service for looking up artifacts that will be used by the ArtifactoryTrigger and any future triggers which would like to use artifact events.

`ArtifactService` should be implemented by anything that needs to look up artifacts to do manual triggering. 

The endpoint for getting an artifact will be called by echo in https://github.com/spinnaker/echo/blob/master/echo-pipelinetriggers/src/main/java/com/netflix/spinnaker/echo/pipelinetriggers/eventhandlers/ManualEventHandler.java#L77

The list for versions will be called by gate and will be used for populating the manual trigger dropdown.

There's a test implementation to demonstrate behavior.


igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/IgorConfig.groovy
```diff
@@ -69,6 +70,11 @@ class IgorConfig extends WebMvcConfigurerAdapter {
         new BuildServices()
     }
 
+    @Bean
```
ezimanyi: Curious about the decision to explicitly create a `@Bean` rather than add a `@Component` annotation on the class...I see examples of both in our codebase, and sometimes wonder which one to do.  (Entirely just a question out of curiosity, not a suggestion to change this.)

igor-web/src/main/java/com/netflix/spinnaker/igor/artifacts/ArtifactController.java
```diff
@@ -0,0 +1,69 @@
...
+import java.util.List;
+
+/**
+ * Controller for supplying version options and artifacts for a trigger that triggers from artifact events.
+ */
+@RestController
+@RequestMapping("/artifacts")
+public class ArtifactController {
+
+  ArtifactServices artifactServices;
```
ezimanyi: Could this be `private final`?

igor-web/src/main/java/com/netflix/spinnaker/igor/artifacts/ArtifactController.java
```diff
@@ -0,0 +1,69 @@
...
+
+  @Autowired
```
jkschneider: `@Autowired` is implied on single non-default constructors.

igor-web/src/test/groovy/com/netflix/spinnaker/igor/artifacts/TestArtifactService.java
```diff
@@ -0,0 +1,65 @@
...
+
+import com.netflix.spinnaker.igor.model.ArtifactServiceProvider;
+import com.netflix.spinnaker.kork.artifacts.model.Artifact;
+import com.netflix.spinnaker.kork.web.exceptions.NotFoundException;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+public class TestArtifactService implements ArtifactService {
```
jkschneider: Can this be relocated to the java source set?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/IgorConfig.groovy
emjburns: For here specifically I followed what you did for `BuildServices` 😄. In general, I'll do `@Component` if that's what everything in the module does. If some are defined with `@Bean` in the config file I'll do that so that everything is in one place. Also, if something is designed to be replaced by a custom component (and therefore needs the `@ConditionalOnMissingBean`) you can only do that with the `@Bean` way. Sometimes if there are `@ConditionalOnProperty` or two services that are turned on/off by one property value I'll put them together in a config so that you can clearly see what the property controls.

Hopefully that helps with how I think about things!

igor-web/src/main/java/com/netflix/spinnaker/igor/artifacts/ArtifactController.java
emjburns: I really dislike magic `@Autowired` (when it is implied like you mention). My strong preference is for a clearly annotated constructor for clarity. In general I don't mind a couple more lines for something that clearly indicates what's going on

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/IgorConfig.groovy
robzienert: > `@Component` (and `@Service` and `@Repository`) are used to auto-detect and auto-configure beans using classpath scanning. There's an implicit one-to-one mapping between the annotated class and the bean (i.e. one bean per class). Control of wiring is quite limited with this approach, since it's purely declarative.

> `@Bean` is used to explicitly declare a single bean, rather than letting Spring do it automatically as above. It decouples the declaration of the bean from the class definition, and lets you create and configure beans exactly how you choose.

https://stackoverflow.com/questions/10604298/spring-component-versus-bean

igor-web/src/main/java/com/netflix/spinnaker/igor/artifacts/ArtifactService.java
```diff
@@ -0,0 +1,43 @@
...
+ * You'll need to implement a list of artifact versions for a given name,
+ * and retrieving a specific artifact by name and version.
+ */
+public interface ArtifactService {
+  ArtifactServiceProvider artifactServiceProvider();
+
+  /**
+   * Used to populate the manual trigger dropdown with options
+   */
+  List<String> getArtifactVersions(String name);
```
robzienert: Perhaps `@Nonnull` these?

igor-web/src/test/groovy/com/netflix/spinnaker/igor/artifacts/TestArtifactService.java
emjburns: Yes! Moving.

igor-web/src/main/java/com/netflix/spinnaker/igor/artifacts/ArtifactServices.java
```diff
@@ -0,0 +1,51 @@
...
+  }
+
+  public List<String> getServiceNames() {
+    return artifactServices.keySet().stream().sorted().collect(Collectors.toList());
+  }
+
+  public List<String> getServiceNames(ArtifactServiceProvider artifactServiceProvider) {
+    return artifactServices.entrySet().stream()
+      .filter(e -> e.getValue() != null)
+      .filter(e -> e.getValue().artifactServiceProvider() == artifactServiceProvider)
```
robzienert: I'd combine these two `filter`s so we're not iterating the stream twice unnecessarily.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/IgorConfig.groovy
ezimanyi: Yup, that makes sense, thanks!

PR #415, opened 2019-04-03
fix(cf): okhttp3 connector now trusts all certificates

jasonjchu:
spinnaker/spinnaker#4238


jkschneider:

PR #418, opened 2019-04-05
fix(travis): Check that logs are complete before parsing them

jervi:
The Travis™ API often returns incomplete logs for a few moments (~30-60 seconds) after a build completes. As we are dependent on the logs for detecting both build artifacts and fetching properties, I've added additional checks that will cause Spinnaker to wait until the logs are ready before trying to parse them. If you thought that there's an endpoint in the Travis API to check if the logs are ready yet, you would be wrong. So we have to check this by actually downloading the logs multiple times until they are complete.

**Pros:** We get rid of a whole bunch of weird, non-deterministic bugs caused by the last few lines of the log to sometimes be cut away. Especially since the last couple of lines is where users usually upload artifacts, our pipelines will now Work More Often™ and Fail Less™ because of missing artifacts and/or properties.

**Cons:** It will take, on average, 30 seconds longer before a completed build triggers a Spinnaker pipeline or a Travis stage completes. And we're probably causing some increased load on Travis since we are asking for logs more often.

**Bonus:** I've added Redis caching for the logs as well, with a TTL of 10 minutes. As soon as the log is ready, it goes into the cache.

### DIFF
--- a/igor-web/src/main/java/com/netflix/spinnaker/igor/travis/client/model/v3/V3Log.java
+++ b/igor-web/src/main/java/com/netflix/spinnaker/igor/travis/client/model/v3/V3Log.java
@@ -60,11 +60,11 @@ public class V3Log {
 
     public boolean isReady() {
         if (logParts == null || logParts.isEmpty()) {
-          return false;
+            return false;
         }
         int numberOfParts = logParts.size() - 1;
-      V3LogPart lastLogPart = logParts.get(numberOfParts);
-      return numberOfParts == lastLogPart.number && lastLogPart.isFinal();
+        V3LogPart lastLogPart = logParts.get(numberOfParts);
+        return numberOfParts == lastLogPart.number && lastLogPart.isFinal();
     }
 
     @JsonIgnoreProperties(ignoreUnknown = true)
--- a/igor-web/src/main/java/com/netflix/spinnaker/igor/travis/client/model/v3/V3Log.java
+++ b/igor-web/src/main/java/com/netflix/spinnaker/igor/travis/client/model/v3/V3Log.java
@@ -71,7 +71,7 @@ public class V3Log {
     static class V3LogPart {
         private String content;
         private Integer number;
-        private boolean _final;
+        private boolean isFinal;
 
         public String getContent() {
             return content;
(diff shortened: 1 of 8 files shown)


jervi:
I will update and fix the tests on Monday.

jervi:
@spinnaker/reviewers PTAL 🙂

igor-web/src/main/java/com/netflix/spinnaker/igor/travis/client/model/v3/V3Log.java
```diff
@@ -58,10 +58,20 @@ public void setLogParts(List<V3LogPart> logParts) {
         this.logParts = logParts;
     }
 
+    public boolean isReady() {
+        if (logParts == null || logParts.isEmpty()) {
+          return false;
+        }
+        int numberOfParts = logParts.size() - 1;
+      V3LogPart lastLogPart = logParts.get(numberOfParts);
```
ajordens: Some indentation shenanigans here

igor-web/src/main/java/com/netflix/spinnaker/igor/travis/client/model/v3/V3Log.java
```diff
@@ -58,10 +58,20 @@ public void setLogParts(List<V3LogPart> logParts) {
...
+      return numberOfParts == lastLogPart.number && lastLogPart.isFinal();
+    }
+
     @JsonIgnoreProperties(ignoreUnknown = true)
     static class V3LogPart {
         private String content;
         private Integer number;
+        private boolean _final;
```
ajordens: Generally I'd prefer the jackson annotation to map a `isFinal` attribute here to `final` in the response payload (assuming that the `_` is because of `final` being a reserved word)

jervi: Better?

igor-web/src/main/java/com/netflix/spinnaker/igor/travis/client/model/v3/V3Log.java
jervi: Fixed

igor-web/src/main/java/com/netflix/spinnaker/igor/travis/client/model/v3/V3Log.java
ajordens: LGTM

PR #419, opened 2019-04-08
fix(concourse): Fix deserialization of put resource metadata

jkschneider:
Put resource ids are buried under two further levels of nesting than get resources:

image

### DIFF
--- a/igor-web/src/main/java/com/netflix/spinnaker/igor/concourse/client/model/Plan.java
+++ b/igor-web/src/main/java/com/netflix/spinnaker/igor/concourse/client/model/Plan.java
@@ -46,12 +46,20 @@ public class Plan {
     private ResourceShape put;
 
     @Nullable
+    private OnSuccess onSuccess;
+
+    @Nullable
     public Resource getResource() {
-      ResourceShape shape = get == null ? put : get;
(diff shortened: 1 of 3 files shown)


jkschneider:
Once put resources are fully deserializable, we can extra metadata from any Concourse put resource, such as we are doing here to extract put resource metadata about build version to a variable:

image

igor-web/src/main/java/com/netflix/spinnaker/igor/concourse/client/model/Plan.java
```diff
@@ -45,13 +45,21 @@
...
 
+    @Nullable
+    private OnSuccess onSuccess;
+
     @Nullable
     public Resource getResource() {
-      ResourceShape shape = get == null ? put : get;
+      String resourceId = id;
+      ResourceShape shape = get;
+      if(shape == null && onSuccess != null) {
```
ajordens: (I hate only calling out minor stylistic bits) one teenie tiny space before `(`

jkschneider: No worries. Appreciate you pointing it out.

PR #424, opened 2019-04-20
 fix(artifactory): Avoid AQL 500 when context root is non-empty

jkschneider:
When the context root is non-empty, the Artifactory client constructs a post request with an extra slash in the path, which in turn causes Artifactory to throw a spurious error about a missing repo key.

```
POST /artifactory//api/search/aql
```

### DIFF
--- a/igor-web/igor-web.gradle
+++ b/igor-web/igor-web.gradle
@@ -63,13 +63,3 @@ dependencies {
     testCompile 'com.github.tomakehurst:wiremock-jre8:2.22.0'
     testCompile 'org.mockito:mockito-core:2.6.8'
 }
-
-configurations.all {
-    resolutionStrategy {
-        force 'org.apache.log4j:log4j:1.2.17'
-        force 'commons-codec:commons-codec:1.7'
-    }
-    exclude group: 'javax.servlet', module: 'servlet-api'
-    exclude group: "org.slf4j", module: "slf4j-log4j12"
-    exclude group: "org.mortbay.jetty", module: "servlet-api"
-}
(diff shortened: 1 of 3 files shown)


igor-web/igor-web.gradle
```diff
@@ -63,13 +63,3 @@ dependencies {
     testCompile 'com.github.tomakehurst:wiremock-jre8:2.22.0'
     testCompile 'org.mockito:mockito-core:2.6.8'
 }
-
-configurations.all {
-    resolutionStrategy {
-        force 'org.apache.log4j:log4j:1.2.17'
```
jkschneider: There is no such group as `org.apache.log4j`. This was always a force that did nothing.

igor-web/igor-web.gradle
```diff
@@ -63,13 +63,3 @@ dependencies {
...
-        force 'commons-codec:commons-codec:1.7'
```
jkschneider: This downgrade breaks the Artifactory monitor.

igor-web/igor-web.gradle
```diff
@@ -63,13 +63,3 @@ dependencies {
...
-    }
-    exclude group: 'javax.servlet', module: 'servlet-api'
```
jkschneider: There is no `servlet-api` dependency in the dependency tree to exclude (of either group), nor an `slf4j-log4j12` dependency.

igor-web/igor-web.gradle
cfieber: maybe this was just future proofing? :grin: 

FWIW the igor boot2 branch has overhauled a bunch of the gradle madness to use a BOM and got rid of the resolutionStrategy as a result

Jammy-Louie:

PR #425, opened 2019-04-22
feat(gcb): Add endpoints for storing and retrieving build status

ezimanyi:
* fix(gcb): Fix deserialization of Build object 

  Jackson's ObjectMapper does not properly deserialize Build objects, as nested properties that extend com.google.api.client.json.GenericJson
  (which itself implements Map) are are deserialized as Maps rather than as the specific type.

  There is probably a way of configuring ObjectMapper to do this properly but I have not (yet) been able to find it. As a workaround, use the Google Cloud supported deserialization classes to deserialize this JSON. In order to keep this workaround localized, make a small class that will handle deserializing Build classes.

  This commit also adds BuildOptions to the build in the integration test; adding this causes the test to fail when using the standard ObjectMapper for deserialization.

* feat(gcb): Add endpoints for storing and retrieving build status 

  Create and endpoint to accept the current status of a build and store it in Redis.  This endpoint is intended to be called by echo when it receives a PubSub message about a build status change.  We take care to look at the timestamp of incoming messages and only update the cache if the incoming information is newer, as PubSub does not guarantee delivery order.

  Create an endpoint to return the current status of the build, as per the cache.  This commit does not implement any fallback to calling the API if the cache does not contain the build; this will be added in a future commit.

* feat(gcb): Create build should return only the build 

  When creating a build, we should return only the build instead of the operation. In general, callers will only be interested in the id of the created build, and it will be easier for callers to just look up the
  "id" of the returned object instead of looking at the deeply nested operation -> metadata -> build -> id field.

### DIFF
--- a/igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildCache.java
+++ b/igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildCache.java
@@ -1,0 +1,107 @@
+/*
+ * Copyright 2019 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.igor.gcb;
+
+import com.netflix.spinnaker.igor.polling.LockService;
+import com.netflix.spinnaker.kork.jedis.RedisClientDelegate;
+import lombok.AccessLevel;
+import lombok.RequiredArgsConstructor;
+
+import java.time.Duration;
+import java.util.Map;
+
+/**
+ * Cache to keep track of the status of Google Cloud builds. In general, this cache will be updated as echo
+ * receives PubSub build notifications and sends them to igor.
+ */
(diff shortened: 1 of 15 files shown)


ezimanyi:
Things to be done in a follow-up PR:
* Add an expiration time to the builds we add to the cache, so that they don't continually grow
* Add a fallback to polling the GCB API if the data in the cache is too old

igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildCache.java
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * Cache to keep track of the status of Google Cloud builds. In general, this cache will be updated as echo
+ * receives PubSub build notifications and sends them to igor.
+ */
+@RequiredArgsConstructor(access=AccessLevel.PRIVATE)
+public class GoogleCloudBuildCache {
+  private final LockService lockService;
+  private final RedisClientDelegate redisClientDelegate;
+  private final String keyPrefix;
+  private final String lockPrefix;
```
maggieneterval: Redis for dummies question, what is the purpose of the lock prefix?

ezimanyi: The lock works by setting a key in redis.  The algorithm is roughly:
* Use the redis "set if not exist" to try to set the key
* If the operation fails, it means the key exists and someone else has the lock; if it succeeds, this process go the lock
* Do the requested work then delete the key

The prefix is the prefix for that actual key we're creating.  We create a key of format `googleCloudBuild.account.buildId` so that the lock is unique to each account/build combo (with the prefix just to have some namespacing).

As we create a `GoogleCloudBuildCache` for each account, the `lockPrefix` there is set to `googleCloudBuild.account` and we just need to append the `buildId` each time we want to get the lock.

maggieneterval: :+1: thanks for the great explanation!

PR #427, opened 2019-04-25
chore(dependencies): update to spring boot 2 and kork BOM dependency management

cfieber:
This updates igor to use the kork BOM as the source of dependency opinions, and updates to spring boot 2.


build.gradle
```diff
@@ -70,15 +55,6 @@ allprojects {
...
-
-    //c&p this because NetflixOss reverts it to 1.7 and ends up getting applied last..
-    project.plugins.withType(JavaBasePlugin) {
-        JavaPluginConvention convention = project.convention.getPlugin(JavaPluginConvention)
-        convention.sourceCompatibility = JavaVersion.VERSION_1_8
-        convention.targetCompatibility = JavaVersion.VERSION_1_8
-    }
 }
 
 defaultTasks 'run'
```
robzienert: I just love how the bom stuff cleans up the main gradle file. 🥓

igor-web/src/main/groovy/com/netflix/spinnaker/igor/Main.groovy
```diff
@@ -48,6 +48,8 @@ class Main extends SpringBootServletInitializer {
     ]
 
     static {
+        System.setProperty("spring.main.allow-bean-definition-overriding", "true");
```
robzienert: I don't think this is necessary anymore - I pulled this out into kork.

build.gradle
cfieber: indeed, is niiice

igor-web/src/main/groovy/com/netflix/spinnaker/igor/Main.groovy
cfieber: good call, saw that in kork

PR #431, opened 2019-04-29
 feat(gcb): Add ttl to build cache entries and fall back to polling

ezimanyi:
* style(gcb): Change indent to 2 spaces 

  The default for igor was recently changed to 2 spaces to match our other microservices; update the 2 gcb files that still have 4-space indents. This commit is whitespace-only.

* refactor(gcb): Extract all interactions with GCB API to a client class 

  Currently GoogleCloudBuildAccount directly generates calls to the GCB API (though uses a GoogleCloudBuildExecutor to actually execute the requests). As we add more complex logic to GoogleCloudBuildAccount and want to add some unit tests, it makes sense to pull the logic that is just generating API calls into a separate class so it can be easily stubbed.

  Create a GoogleCloudBuildClient which will expose GCB API operations and execute them for the consumer.

* refactor(gcb): Standardize return values in GoogleCloudBuildAccount 

  We'll standardize on having the account class accept and return GCB objects rather than serialized objects or maps.

* feat(gcb): Add ttl to build cache entries and fall back to polling 

  In order to handle cases where a pubsub message is not delivered, we'd like to fall back to polling if we haven't received an update on a build in some amount of time.

  Configure this by adding a TTL on all entries in the cache; we'll set a shorter TTL for in-progress builds (as we expect an update) and a longer one for completed builds (which should never have another status change). If we check the cache and there is no build there, fall back to an external poll, and store that result for the same amount of time.

### DIFF
--- a/igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildCache.java
+++ b/igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildCache.java
@@ -20,6 +20,7 @@ import com.netflix.spinnaker.igor.polling.LockService;
 import com.netflix.spinnaker.kork.jedis.RedisClientDelegate;
 import lombok.AccessLevel;
 import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
 
 import java.time.Duration;
 import java.util.Map;
@@ -29,6 +30,7 @@ import java.util.Map;
  * receives PubSub build notifications and sends them to igor.
  */
 @RequiredArgsConstructor(access=AccessLevel.PRIVATE)
+@Slf4j
 public class GoogleCloudBuildCache {
   private static final int inProgressTtlSeconds = 60 * 10;
   private static final int completedTtlSeconds = 60 * 60 * 24;
@@ -82,6 +84,7 @@ public class GoogleCloudBuildCache {
         return inProgressTtlSeconds;
       }
     } catch (IllegalArgumentException e) {
+      log.warn("Received unknown Google Cloud Build Status: {}", statusString);
       return inProgressTtlSeconds;
     }
   }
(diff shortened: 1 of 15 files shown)


igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildCache.java
```diff
@@ -62,12 +65,27 @@ private void internalUpdateBuild(String buildId, String status, String build) {
...
+  private int getTtlSeconds(String statusString) {
+    try {
+      GoogleCloudBuildStatus status = GoogleCloudBuildStatus.valueOf(statusString);
+      if (status.isComplete()) {
+        return completedTtlSeconds;
+      } else {
+        return inProgressTtlSeconds;
+      }
+    } catch (IllegalArgumentException e) {
+      return inProgressTtlSeconds;
```
jeyrschabu: Not a big fan of swallowing exceptions. Is this something that we can log?

igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildClient.java
```diff
@@ -0,0 +1,55 @@
...
+    private final String applicationName;
+
+    public GoogleCloudBuildClient create(GoogleCredential credential, String projectId) {
+      CloudBuild cloudBuild = cloudBuildFactory.getCloudBuild(credential, applicationName);
+      return new GoogleCloudBuildClient(projectId, cloudBuild, executor);
+    }
+  }
+
+  public Operation createBuild(Build build) {
+    return executor.execute(() -> cloudBuild.projects().builds().create(projectId, build));
```
jeyrschabu: This would execute in a different thread right?

igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildCache.java
ezimanyi: We're swallowing a very targeted exception, where the status on the received GCB build isn't one we know about---as if GCB adds a new status, I'd rather the failure mode be less efficient caching than actual user-facing breakage.

It does seem reasonable to log this since if it happens we'd want to update igor to know of the new status, so I'll add that in.

igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildClient.java
ezimanyi: No, this is operating in the same thread, just using a helper class to handle actually sending the request to the Google API.

PR #433, opened 2019-04-29
feat(gcb): Add endpoint to fetch build artifacts

ezimanyi:
* feat(gcb): Add support for parsing GCS paths 

  In order to fetch the list of artifacts produced by a build, we'll need to fetch the manifest from GCS. This will involve translating a GCS path into a bucket/object/version, which there is no helper function in the GCS library for; write a small class to do this and test it.

* feat(gcb): Add GCS object to GCB account 

  As Google Cloud Build stores a manifest of produced artifacts in Google Cloud Storage, a GCB account will need to be able to talk to GCS in order to return information about produced artifacts.

  This allows us to better encapsulate information about produced artifacts rather than having the GCB account simply return a path to the manfest and require the caller to parse any artifacts (which are in a GCB-specific format).

* feat(gcb): Add endpoint to fetch build artifacts 

  The endpoint should return a list of Spinnaker artifacts representing artifacts produced by the build.  A build can produce either docker containers or arbitrary files in Google Cloud Storage.

  Docker containers are directly included in the build output, so we can read the relevant fields and construct an artifact. GCS files are not included directly in the output; the output contains a GCS path to a manifest of produced files. We need to fetch and parse this manifest in order to generate the produced artifacts.

* feat(gcb): Add endpoint for extracting artifacts from build

  In some cases the consumer already has a build from which they'd like to extract artifacts; allow the consumer to send this build to igor rather than just the id (which could have igor fetch stale data).

### DIFF
--- a/igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/model/GoogleCloudStorageObject.java
+++ b/igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/model/GoogleCloudStorageObject.java
@@ -1,0 +1,89 @@
+/*
+ * Copyright 2019 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License")
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.igor.gcb.model;
+
+import lombok.AccessLevel;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+@Getter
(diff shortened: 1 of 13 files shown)


igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/model/GoogleCloudStorageObject.java
```diff
@@ -0,0 +1,89 @@
...
+
+package com.netflix.spinnaker.igor.gcb.model;
+
+import lombok.AccessLevel;
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+
+@Getter
+@RequiredArgsConstructor(access = AccessLevel.PRIVATE)
+public class GoogleCloudStorageObject {
```
jtk54: Kind of surprised there's no helper class for this in the client lib.

ezimanyi: Yeah, I think it's because the client is just a generated client for the REST APIs which only understand `bucket` and `object` syntax, whereas the `gs://bucket/object` is handled by `gsutil`.  There is a newer, more idiomatic library to supersede the generated one, but we're using the generated one everywhere else (and the new one doesn't support GCB yet)...maybe that one supports this, but I didn't want to bring it in just for this, so I didn't check.

PR #448, opened 2019-05-29
fix(travis): /repos endpoint is not paginated anymore

jervi:
In some cases this has led to strange results as the same list of repos has been returned multiple times


jervi:
@spinnaker/reviewers PTAL

jervi:
Not needed anymore when #467 is merged

PR #449, opened 2019-05-29
feat(travis): Support for log_complete from Travis API

jervi:
Travis added a new property in version 2.2.9 (also available on travis-ci.org) that tells us if the log is complete or not. This allows us to be more efficient in the way we fetch artifacts and properties from Travis. It is not backwards compatible, so we have to keep the old way of doing it around for a while.

Each Travis master will be queried at Igor startup for some simple feature detection to determine wether to use the legacy method or the new method.

A few interfaces has been changed to use GenericBuild instead of `buildNumber`. This allows for some reduction in complexity for a few of the providers that don't use `buildNumber` internally, as we don't have to query the master for `buildId` multiple times.

Also contains some spring cleaning of the TravisClient and the TravisService, it contained a lot of unused methods.

*Note:* Even with this new `log_complete` property, we need to try to download the log for new builds, otherwise the triggers will be triggered very late. The reason is that the `log_complete` will be false for a while even when the log is actually complete, it just isn't concatenated in the API yet. We instead do the concatenation manually after checking that we have all the parts and that the last part is the final one. If Travis fixes this in the future, we can remove this code path.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
@@ -74,7 +74,7 @@ class BuildController {
             return null
 
         try {
-            build.genericGitRevisions = buildService.getGenericGitRevisions(job, buildNumber)
+            build.genericGitRevisions = buildService.getGenericGitRevisions(job, build)
         } catch (Exception e) {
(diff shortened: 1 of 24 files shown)


jervi:
@spinnaker/reviewers PTAL

jervi:
The changes in this PR are included in #467, but I think we should probably merge this on its own, given the size of these PR's.

PR #454, opened 2019-06-05
fix(boot2): Allow paths with URL-encoded slashes

avram:
Spring Boot 2 tightened up restrictions, but CI job names may legitimately include encoded slashes

See https://github.com/spinnaker/spinnaker/issues/4477

https://spinnakerteam.slack.com/archives/C091CCWRJ/p1559269913055000


avram:
I would love to actually add a test for this, but none of the tests I looked at seemed to actually run requests through the Spring request handling logic (which is what is changed here) and I was hoping to limit the amount of new test boilerplate for this fix.

It would also be nice to just relax this for the specific endpoints that need it, but it's not clear that we can do that with `StrictHttpFirewall`.

emjburns:
LGTM

PR #464, opened 2019-06-18
feat(igor-bom): publish igor-bom

marchello2000:
This simplifies dependency resolution downstream

### DIFF
--- a/igor-bom/igor-bom.gradle
+++ b/igor-bom/igor-bom.gradle
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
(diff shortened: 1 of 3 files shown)


igor-bom/igor-bom.gradle
```diff
@@ -0,0 +1,49 @@
...
+      }
+    }
+  }
+}
+
+dependencies {
+  api(platform("com.netflix.spinnaker.kork:kork-bom:$korkVersion"))
+
+  constraints {
+    rootProject
```
cfieber: think this needs re-exported similar to [this comment](https://github.com/spinnaker/gate/pull/829#discussion_r294920252) on your gate PR

PR #467, opened 2019-06-20
feat(travis): Refactor and simplify Travis polling

jervi:
Previously, the polling (using the Travis v2 API) asked Travis for ~50 repos with recently built builds, and then, _for each repo_, asked for the 5 most recent builds. In other words, each polling cycle hits Travis with ~51 potentially heavy API calls (plus more to fetch logs). This commit brings that number down to a significantly lower number, using the v3 API, and also removes a lot of code that belongs to the v2 API.~~or was just band-aid for the quirks in the v2 API. One of those quirks was that some builds sometimes disappeared from the API as soon as they were built. I'm currently trying to verify if this bug is still present in v3, so hold off merging for now. But given the size of the PR, I wanted to open it to start collecting some feedback.~~

Igor will still be keeping track of builds in-flight, because some long running jobs can fall outside the query window on busy Travis installations. For these builds, we fall back to querying them directly.

I believe the logic is sound now. Our internal testing indicates that the new logic is more correct than the old (doesn't suddenly trigger builds days later for instance).


jervi:
@spinnaker/reviewers PTAL

PR #472, opened 2019-06-28
chore(gradle): Upgrade the Spinnaker gradle plugin

plumpy:
Just making sure the new IDEA changes to the plugins don't cause problems if the IDEA files don't exist in the repo.


plumpy:
cc @ezimanyi

PR #492, opened 2019-08-23
chore(docs): document Igor's behavior and features in README.md

dreynaud:


README.md
```diff
@@ -1,15 +1,120 @@
...
+
+- periodically get a list of items from an external resource (e.g. builds on a Jenkins master)
+- compare that list against their own persisted cache of items (the difference is called _delta size_)
+- send an `echo` event for each new item
+- caches the new list of items
+
+Features:
+
+- *locking*: pollers can optionally acquire a distributed lock in the storage system before attempting to complete a polling cycle. This makes it possible to run Igor in a high-availability configuration and scale it horizontally.
+- *safe guards*: abnormally large delta sizes can indicate a problem (e.g. lost or corrupt cache data) and cause downstream issues. If a polling cycle results in a delta size above the threshold
```
marchello2000: i am sure you wanted to say something else after `above the threshold`

README.md
```diff
@@ -1,15 +1,120 @@
...
+Relevant properties:
+
+| *Property* | *Default value* | *Description* |
+| --- | --- | --- |
+| `spinnaker.build.pollInterval` | `60` | Interval in seconds between polling cycles |
+| `spinnaker.pollingSafeguard.itemUpperThreshold` | `1000` | Defines the upper threshold for number of new items before a cache update cycle will be rejected | `locking.enabled` | `false` | Enables distributed locking so that Igor can run on multiple nodes without interference |
+
+Relevant metrics:
+
+| *Property* | *Type* | *Description* |
```
marchello2000: probably
```suggestion
| *Metric* | *Type* | *Description* |
```

README.md
```diff
@@ -1,15 +1,120 @@
...
+| --- | --- | --- |
+| `pollingMonitor.newItems` | gauge | represents the number of new items cached by a given monitor during a polling cycle |
+| `pollingMonitor.itemsOverThreshold` | gauge | 0 if deltaSize < threshold, deltaSize otherwise |
+| `pollingMonitor.pollTiming` | timer | published for every polling cycle with the duration it took to complete |
```
marchello2000: do you think it's worth mentioning this event is fired with the name of the monitor as a tag `monitor`

README.md
marchello2000: Also, maybe worth mentioning the `admin` EP to bypass this safeguard?
`admin/pollers/fastforward/{monitorName}`

README.md
```diff
@@ -1,15 +1,120 @@
-# Igor
-[![Build Status](https://api.travis-ci.org/spinnaker/igor.svg?branch=master)](https://travis-ci.org/spinnaker/igor)
+Igor is a service that provides a single point of integration with CI and SCM services for Spinnaker.
```
ezimanyi:
```suggestion
Igor is a service that provides a single point of integration with Continuous Integration (CI) and Source Control Management (SCM) services for Spinnaker.
```

README.md
```diff
@@ -1,15 +1,120 @@
...
 
-Igor provides a single point of integration with Jenkins, Travis and Git repositories ( BitBucket, Stash, and Github ) within Spinnaker.
+# Common Polling Architecture
+
+Igor runs a number of pollers that all share the same common architecture. At a high level, they all:
+
+- periodically get a list of items from an external resource (e.g. builds on a Jenkins master)
+- compare that list against their own persisted cache of items (the difference is called _delta size_)
+- send an `echo` event for each new item
+- caches the new list of items
```
ezimanyi:
```suggestion
- cache the new list of items
```

README.md
ezimanyi:
```suggestion
- *safeguards*: abnormally large delta sizes can indicate a problem (e.g. lost or corrupt cache data) and cause downstream issues. If a polling cycle results in a delta size above the threshold
```

README.md
```diff
@@ -1,15 +1,120 @@
...
+# Integration with CI services
+
+The following CI services are supported:
+
+- Jenkins
+- Gitlab CI
+- Wercker
+- Travis
+- Artifactory
+- Concourse
```
ezimanyi: Google Cloud Build (GCB) is also supported.  (Though it is a bit different in that it doesn't poll GCB and requires setting up pubsub subscriptions as per [here](https://www.spinnaker.io/setup/ci/gcb/)).  Also maybe it makes sense to alphabetize the list?

README.md
dreynaud: Haha good catch! I was probably actually wondering what does happen we're above the threshold and then got sidetracked, refrigerator

README.md
dreynaud: Ah, I didn't know about GCB. I was literally looking for subclasses of `CommonPollingMonitor` :)

Good point about the ordering 👍

dreynaud:
Thanks a bunch for the input @marchello2000 and @ezimanyi, all good points!

PR #506, opened 2019-09-06
fix(nexus): add nexus name endpoint and artifact location

claymccoy:


claymccoy:
@spinnakerbot cherry-pick 1.16

PR #514, opened 2019-09-23
chore(travis): Remove old build cache migrator

jervi:
This migrator has been in place since October 2017, and is no longer needed. Also it was a bit misleading. The `initialize` name suggests it is run on bean creation, but it actually ran every polling cycle, which seems a bit excess for a migrator. Since it was only used by the Travis integration, I just removed the whole thing.


jervi:
@gardleopard FYI

PR #523, opened 2019-10-22
feat(gcb): Add permissioning to Google Cloud Build accounts

afcastano:
Added permissioning to Google Cloud Build accounts.  
It solves https://github.com/spinnaker/spinnaker/issues/4814

Protects the gcb accounts so that only defined roles can READ and WRITE.

- To be able to configure a pipeline with the gcb account, you need READ permissions.
- To be able to submit a CloudBuild job using the account, you need WRITE permissions.


afcastano:
Hi @ezimanyi , this is the PR as per our discussion in spinnaker/spinnaker#4814 . Please let me know what do you think.

I did a bit of manual testing in my local environment and seems working all right.

afcastano:
Great, thanks @ezimanyi, the minor fix is done. =)

PR #525, opened 2019-10-22
fix(concourse): Fix caching of concourse build events

donnyyung:
Builds should be added to cache correctly now and the admin fastforward command should work
It solves https://github.com/spinnaker/spinnaker/issues/5047

I ran into this problem because we had over 1000 builds on our concourse instance, so we were running into `cache exceeds upper threshold (1000)`.

Upon trying to fix this, I realized the builds were never being compared to what was in the cache until the commitDelta function, which unfortunately we don't get to unless the threshold is less than 1000.  

Note: This is my first PR to a spinnaker project and my first time coding in java so open to any and all suggestions!

### DIFF
--- a/igor-web/src/main/java/com/netflix/spinnaker/igor/concourse/ConcourseBuildMonitor.java
+++ b/igor-web/src/main/java/com/netflix/spinnaker/igor/concourse/ConcourseBuildMonitor.java
@@ -124,7 +124,13 @@ public class ConcourseBuildMonitor
     List<GenericBuild> genericBuilds =
         builds.stream()
             .map(build -> concourseService.getGenericBuild(jobPath, build, false))
+            .filter(b -> !cache.getEventPosted(host, job, cursor, b.getNumber()))
             .collect(Collectors.toList());
+
+    if (genericBuilds.size() == 0) {
+      cache.setLastPollCycleTimestamp(host, job, cursor);
+      return null;
+    }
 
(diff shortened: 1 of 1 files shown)


igor-web/src/main/java/com/netflix/spinnaker/igor/concourse/ConcourseBuildMonitor.java
```diff
@@ -124,8 +124,13 @@ private JobDelta jobDelta(ConcourseProperties.Host host, Job job) {
     List<GenericBuild> genericBuilds =
         builds.stream()
             .map(build -> concourseService.getGenericBuild(jobPath, build, false))
+            .filter(b -> !cache.getEventPosted(host, job, cursor, b.getNumber()))
```
asher: We still want to run `cache.setLastPollCycleTimestamp()` for jobs without uncommitted successful builds, which filtering here prevents.

donnyyung: Ah okay. I should be able to add a `cache.setLastPollCycleTimestamp()` in the `if` right after this then right?  

I believe we need the filter + return null there when there are no uncommitted successful builds or else the JobPollingDelta list will always be the # of concourse jobs, correct?  
Just want to make sure I'm understanding the problem.

donnyyung:
@spinnakerbot cherry-pick 1.16

PR #535, opened 2019-10-25
feat(keel): send keel docker events if enabled

emjburns:
In order to react to new docker containers in `keel` we need to know when there are new containers. This PR adds functionality to emit events to `keel` with the same criteria as sending new events to echo. 

This won't kick in unless you have keel enabled. The format for the docker event we send is subject to change, but I think this is a decent first start 🤷‍♀ we shall see.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerMonitor.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerMonitor.groovy
@@ -195,18 +195,17 @@ class DockerMonitor extends CommonPollingMonitor<ImageDelta, DockerPollingDelta>
 
         if (keelService.isPresent()) {
           String imageReference = image.repository + ":" + image.tag
-          Artifact artifact = new Artifact(
-            "DOCKER_IMAGE",
-            false,
-            image.repository,
-            image.tag,
-            image.account,
-            imageId,
-            [fullname: imageReference, registry: image.account, tag: image.tag],
-            null,
-            image.registry,
-            null
-          )
+          Artifact artifact = Artifact.builder()
+            .type("DOCKER")
+            .customKind(false)
+            .name(image.repository)
+            .version(image.tag)
+            .location(image.account)
+            .reference(imageId)
+            .metadata([fullname: imageReference, registry: image.account, tag: image.tag],)
+            .provenance(image.registry)
+            .build()
+
           Map artifactEvent = [
             payload: [artifacts: [artifact], details: [:]],
             eventName: "spinnaker_artifacts_docker"
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerMonitor.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerMonitor.groovy
@@ -26,11 +26,13 @@ import com.netflix.spinnaker.igor.docker.model.DockerRegistryAccounts
 import com.netflix.spinnaker.igor.docker.service.TaggedImage
 import com.netflix.spinnaker.igor.history.EchoService
 import com.netflix.spinnaker.igor.history.model.DockerEvent
+import com.netflix.spinnaker.igor.keel.KeelService
 import com.netflix.spinnaker.igor.polling.CommonPollingMonitor
 import com.netflix.spinnaker.igor.polling.DeltaItem
 import com.netflix.spinnaker.igor.polling.LockService
 import com.netflix.spinnaker.igor.polling.PollContext
 import com.netflix.spinnaker.igor.polling.PollingDelta
+import com.netflix.spinnaker.kork.artifacts.model.Artifact
 import com.netflix.spinnaker.security.AuthenticatedRequest
 import org.springframework.beans.factory.annotation.Autowired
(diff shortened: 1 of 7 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerMonitor.groovy
```diff
@@ -187,6 +192,27 @@ class DockerMonitor extends CommonPollingMonitor<ImageDelta, DockerPollingDelta>
...
+            [fullname: imageReference, registry: image.account, tag: image.tag],
+            null,
+            image.registry,
+            null
+          )
+          Map artifactEvent = [
+            payload: [artifacts: [artifact], details: [:]],
+            eventName: "spinnaker_artifacts_docker"
+          ]
+          AuthenticatedRequest.allowAnonymous { keelService.get().sendArtifactEvent(artifactEvent) }
```
luispollo: Is the `allowAnonymous` bit common practice/OK?

emjburns: It's common for service to service calls, i.e. we're not doing this call on behalf of a user so we have no user auth info to pass on.

luispollo:
P.S. Sorry if this is a silly question, but if we were already sending an event to echo, and echo centralizes most event handling/coordination with other services, shouldn’t we just extend the event handler in echo to call keel?

emjburns:
@luispollo that's a good question. Since the information is coming from `igor`, I didn't want to have it go through an extra service and have an extra network hop to get to `keel`. I thought it would be simpler to have it come from the "source" (well, the source that's calculating deltas). @asher do you have thoughts on this?

luispollo:
BTW @emjburns I noticed that [this echo PR](https://github.com/spinnaker/echo/pull/469/files) from earlier this year introduced echo -> keel integration for artifact events. How does that relate/differ to what we're trying to accomplish here?

igor-web/src/main/java/com/netflix/spinnaker/igor/config/KeelConfig.java
```diff
@@ -0,0 +1,57 @@
...
+
+  @Bean
+  RestAdapter.LogLevel retrofitLogLevel(
+      @Value("${retrofit.log-level:BASIC}") String retrofitLogLevel) {
+    return RestAdapter.LogLevel.valueOf(retrofitLogLevel);
+  }
+
+  @Bean
+  KeelService keelService(
+      OkHttpClientConfiguration okHttpClientConfig,
```
ezimanyi: It looks like `OkHttpClientConfiguration` is deprecated in favor of `OkHttp3ClientConfiguration` so new usages should probably use the new one.  (Though I'm not too familiar with why it was deprecated and what's involved in migrating.)

igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerMonitor.groovy
```diff
@@ -187,6 +192,27 @@ class DockerMonitor extends CommonPollingMonitor<ImageDelta, DockerPollingDelta>
             account: image.account,
           ), artifact: dockerArtifact))
         }
+
+        if (keelService.isPresent()) {
+          String imageReference = image.repository + ":" + image.tag
+          Artifact artifact = new Artifact(
```
ezimanyi: I'd encourage using `Artifact.Builder` here instead of calling the all-args constructor; otherwise any additional (optional) fields that get added to artifact in the future will break this call when kork is bumped.  (Maybe we should even make that constructor private assuming it hasn't been used too many places already.)

igor-web/src/main/java/com/netflix/spinnaker/igor/config/KeelConfig.java
robfletcher: It's part of the shift between Retrofit 1 and Retrofit 2. I had a PR open that split those things in Kork more clearly. I will re-open it and get it merged.

emjburns:
(answered Luis's questions offline)

igor-web/src/main/java/com/netflix/spinnaker/igor/config/KeelConfig.java
emjburns: @ezimanyi I'll admit, I couldn't find a good example of using `OkHttp3ClientConfiguration` so I was lazy. @robfletcher do you have a good example of how to do things the recommended way?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerMonitor.groovy
emjburns: Oh yes that will be much better. Overlooked the builder. Will change! Thanks.

emjburns: I think it would be worth making that constructor private

igor-web/src/main/java/com/netflix/spinnaker/igor/config/KeelConfig.java
emjburns: Found one! Updated this.

PR #539, opened 2019-11-04
feat(jenkins): Add retry logic on getting build artifact

evansb:
This is one way to fix https://github.com/spinnaker/spinnaker/issues/5055
Another way is to just throw the exception without adding retry logic so that the retry logic belongs in Orca. However I noticed that the other operation `getArtifactPathFromBuild` already has a retry logic so I am going with this approach for consistency.

This PR add automatic retry logic to Jenkins' `getPropertyFile` similar to `getGitDetails` (sans Hystrix command).

Before this PR, if Jenkins return transient failure on `getPropertyFile` it will not retried and the Jenkins stage will immediately fail. User will have to restart the stage just to get the build artifact.

After this PR, Igor will retry up to 5 times with 2 seconds non-exponential backoff if Jenkins return 5xx code.

### DIFF
--- a/igor-web/src/test/groovy/com/netflix/spinnaker/igor/jenkins/service/JenkinsServiceSpec.groovy
+++ b/igor-web/src/test/groovy/com/netflix/spinnaker/igor/jenkins/service/JenkinsServiceSpec.groovy
@@ -22,10 +22,18 @@ import com.netflix.spinnaker.igor.build.model.GenericGitRevision
 import com.netflix.spinnaker.igor.config.JenkinsConfig
 import com.netflix.spinnaker.igor.config.JenkinsProperties
 import com.netflix.spinnaker.igor.jenkins.client.JenkinsClient
+import com.netflix.spinnaker.igor.jenkins.client.model.Build
+import com.netflix.spinnaker.igor.jenkins.client.model.BuildArtifact
 import com.netflix.spinnaker.igor.jenkins.client.model.BuildsList
 import com.netflix.spinnaker.igor.jenkins.client.model.Project
+import com.netflix.spinnaker.kork.exceptions.SpinnakerException
 import com.squareup.okhttp.mockwebserver.MockResponse
 import com.squareup.okhttp.mockwebserver.MockWebServer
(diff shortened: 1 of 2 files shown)


igor-web/src/test/groovy/com/netflix/spinnaker/igor/jenkins/service/JenkinsServiceSpec.groovy
```diff
@@ -683,4 +691,62 @@ class JenkinsServiceSpec extends Specification {
...
+        null
+      )
+      def propertyFile = new Response("http://my.jenkins.net", 200, "", [], new TypedString("a=b"))
+
+      when:
+      service.getBuildProperties(jobName, build.genericBuild(jobName), artifact.fileName)
+
+      then:
+      1 * client.getBuild(jobName, buildNumber) >> build
+      client.getPropertyFile(jobName, buildNumber, artifact.fileName) >> { throw badGatewayError } >> propertyFile
```
emjburns: Does this actually verify that the code retries and ends up with the correct property file? It looks like this line is just providing the mock values, since the return value in the `when` block is not getting assigned to anything/checked that the response is correct.

emjburns: Although if it was called multiple times but there was no answer for the mock, would the framework complain?

evansb: I'm quite new, I think you are right, I'm quite new to Spock but from the docs it seems that I should not omit the cardinality (`2 *`) or it will just be a naked mock expression.

Added a check to the property file as well and verified that both actually do verify

emjburns: I always struggle with Spock, so I thought I'd ask :) Thanks for updating the tests! The behavior is much clearer now.

PR #545, opened 2019-11-06
feat(gcb): Allow invoking existing GCB triggers

afcastano:
Added ability to call existing GCB triggers.
This is part of: https://github.com/spinnaker/spinnaker/issues/5076

Added two new endpoints to `igor` to:
* List existing triggers available for the gcb account.
* Call a trigger given its id. 

Calling a trigger will return a Build object, which then can be used to monitor the build using the existing functionality in `orca`.

### DIFF
--- a/igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildAccount.java
+++ b/igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildAccount.java
@@ -17,7 +17,10 @@
 package com.netflix.spinnaker.igor.gcb;
 
 import com.google.api.services.cloudbuild.v1.model.Build;
+import com.google.api.services.cloudbuild.v1.model.BuildTrigger;
+import com.google.api.services.cloudbuild.v1.model.ListBuildTriggersResponse;
 import com.google.api.services.cloudbuild.v1.model.Operation;
+import com.google.api.services.cloudbuild.v1.model.RepoSource;
 import com.netflix.spinnaker.kork.artifacts.model.Artifact;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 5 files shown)


afcastano:
Hi @ezimanyi , this is the first PR for enabling calling GCB triggers. I'll start working in the `orca` changes, but it'll be good if you can take a look at this one in the mean time.

I did some local tests using `curl` and it seems working all right.

My main question would be whether I need to implement any type of cache for the trigger list, as it takes around one second to fetch the list and return it to the caller.

Anyway, let me know what do you think and thanks for helping me.
Cheers,

igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildController.java
```diff
@@ -86,4 +88,25 @@ Build getBuild(@PathVariable String account, @PathVariable String buildId) {
...
   }
+
+  @RequestMapping(value = "/triggers/{account}", method = RequestMethod.GET)
+  @PreAuthorize("hasPermission(#account, 'BUILD_SERVICE', 'READ')")
+  List<BuildTrigger> listTriggers(@PathVariable String account) {
+    return googleCloudBuildAccountRepository.getGoogleCloudBuild(account).listTriggers();
+  }
+
+  @RequestMapping(
+      value = "/triggers/run/{account}/{triggerId}",
```
ezimanyi: I think this mapping might be better as `/triggers/{account}/{triggerId}/run` to keep the resource identifier (`/triggers/{account}/{triggerid}`) together and move the custom verb to the end.  (This would also keep it consistent with how we mapped the `getArtifacts` custom action above, by adding the verb at the end of the resource.)

igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildAccount.java
```diff
@@ -71,6 +74,22 @@ public Build getBuild(String buildId) {
     return googleCloudBuildParser.parse(buildString, Build.class);
   }
 
+  public List<BuildTrigger> listTriggers() {
```
ezimanyi: Would like to see a simple test in `GoogleCloudBuildAccountSpec` for `listTriggers` and `runTrigger`...really just ensuring that we pass the data along to the `client` and return the response.  The first test (`createBuild creates a build and returns the result`) should be a good example.

igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildController.java
afcastano: Makes sense, will implement it that way.

igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildAccount.java
afcastano: I missed that one. Didn't search for references in groovy classes. Will add some tests. Thanks!

afcastano:
@ezimanyi I have implemented the changes. Didn't squash the commits just to make it easier for you to review. Please let me know.
Cheers, and thanks again.

PR #556, opened 2019-12-03
feat(Artifactory Trigger): Add helm repo artifacts support to artifac…

sirigithub:
This feature enhances the Artifactory Trigger to support helm artifacts. The feature works similar to the existing maven artifacts where a search is configured for a helm repo type. The enhancement introduces one new parameter in the search configuration : 

repoType: This field identifies the artifactory repository type  Maven, Helm etc

(Note there is a reference to a PR in Halyard to support new variables in search)

The search configuration looks like:
``` yaml
searches:
    -name: test-helm
     permissions: {}
     baseUrl: https://artifactory-url/artifactory
     repo: test-helm-local
     repoType: helm
     username: username
     password: ..
   -name: test-maven
    permissions: {}
    baseUrl: https://artifactory-url/artifactory
    repo: test-maven-local
    groupId: com.mycompany
    repoType: maven
    username:
```
We extended the enum ArtifactoryRepositoryType to include helm:
```java
public enum ArtifactoryRepositoryType {
  MAVEN("Maven", ".pom"),
  HELM("HELM", ".tgz");

  private final String repoTypeString;
  private final String artifactExtension;
}
```
The ArtifactorySearch.java has a new utility method that returns the artifactExtension based on the repoType configured in the search
``` java
public String getArtifactExtension() {
    return repoType.getArtifactExtension();
  }
```
The aql used to query artifactory in ArtifactoryBuildMonitor.java has been modified to match the repo and file type based on the repo type configured in search
```java
String aqlQuery =
                  "items.find({"
                      + "\"repo\":\""
                      + search.getRepo()
                      + "\","
                      + modified
                      + ","
                      + "\"path\":{\"$match\":\""
                      + (search.getGroupId() == null
                          ? ""
                          : search.getGroupId().replace('.', '/') + "/")
                      + "*\"},"
                      + "\"name\": {\"$match\":\""
                      + "*."
                      + search.getSearchPattern()
``` 
In order to match a helm artifact found during a poll , the method toMatchableArtifact in ArtifactoryItem.java has been modified to return a Artifact fortype Helm:
```java
case HELM:
        String filePath = null;
        if (baseUrl != null) {
          filePath = baseUrl + "/webapp/#/artifacts/browse/tree/General/" + repo + "/" + name;
        }

        String helmVersion = getHelmFileVersion(name);
        Artifact.ArtifactBuilder artifactBuilderHelm =
            Artifact.builder()
                .type("helm/file")
                .reference("")
                .name(name)
                .version(helmVersion)
                .provenance(repo)
                .location(filePath);
        return artifactBuilderHelm.build();
``` 
A new class ArtifactoryRepositoryTypeConverter has been introduced to convert the repoType String to a Enum
```java
public class ArtifactoryRepositoryTypeConverter
    implements Converter<String, ArtifactoryRepositoryType> {

  @Override
  public ArtifactoryRepositoryType convert(String source) {
    return ArtifactoryRepositoryType.valueOf(source.toUpperCase());
  }
}
```

We did not include logic to look build modules in Artifactory for helm charts as most of the deployment use cases we have seen are done via curl. If this is not a accurate assumption, we can rework the code.

Referenced feature request: https://github.com/spinnaker/spinnaker/issues/5074
Referenced Halyard PR to support new search fields in ArtifactorySearch.java : https://github.com/spinnaker/halyard/pull/1472

### DIFF
--- a/igor-monitor-artifactory/src/main/java/com/netflix/spinnaker/igor/artifactory/ArtifactoryBuildMonitor.java
+++ b/igor-monitor-artifactory/src/main/java/com/netflix/spinnaker/igor/artifactory/ArtifactoryBuildMonitor.java
@@ -135,11 +135,11 @@ public class ArtifactoryBuildMonitor
                           ? ""
                           : search.getGroupId().replace('.', '/') + "/")
                       + "*\"},"
-                      + "\"name\": {\"$match\":\""
-                      + "*."
-                      + search.getSearchPattern()
+                      + "\"name\": {\"$match\":\"*"
+                      + search.getArtifactExtension()
                       + "\"}"
                       + "}).include(\"path\",\"repo\",\"name\", \"artifact.module.build\")";
+
               ArtifactoryRequest aqlRequest =
                   new ArtifactoryRequestImpl()
                       .method(ArtifactoryRequest.Method.POST)
@@ -159,7 +159,7 @@ public class ArtifactoryBuildMonitor
                       search.getPartitionName(),
                       Collections.singletonList(
                           new ArtifactDelta(
-                              System.currentTimeMillis(), search.getRepositoryType(), results)));
+                              System.currentTimeMillis(), search.getRepoType(), results)));
                 }
 
                 log.warn(
--- a/igor-monitor-artifactory/src/main/java/com/netflix/spinnaker/igor/artifactory/ArtifactoryBuildMonitor.java
+++ b/igor-monitor-artifactory/src/main/java/com/netflix/spinnaker/igor/artifactory/ArtifactoryBuildMonitor.java
@@ -136,9 +136,10 @@ public class ArtifactoryBuildMonitor
                           : search.getGroupId().replace('.', '/') + "/")
                       + "*\"},"
                       + "\"name\": {\"$match\":\""
-                      + "*.pom\"}"
+                      + "*."
+                      + search.getSearchPattern()
+                      + "\"}"
                       + "}).include(\"path\",\"repo\",\"name\", \"artifact.module.build\")";
-
               ArtifactoryRequest aqlRequest =
                   new ArtifactoryRequestImpl()
                       .method(ArtifactoryRequest.Method.POST)
@@ -158,7 +159,7 @@ public class ArtifactoryBuildMonitor
                       search.getPartitionName(),
                       Collections.singletonList(
                           new ArtifactDelta(
-                              System.currentTimeMillis(), search.getRepoType(), results)));
+                              System.currentTimeMillis(), search.getRepositoryType(), results)));
                 }
 
                 log.warn(
--- a/igor-monitor-artifactory/src/main/java/com/netflix/spinnaker/igor/artifactory/ArtifactoryRepositoryTypeConverter.java
+++ b/igor-monitor-artifactory/src/main/java/com/netflix/spinnaker/igor/artifactory/ArtifactoryRepositoryTypeConverter.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright 2019 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.igor.artifactory;
+
+import com.netflix.spinnaker.igor.artifactory.model.ArtifactoryRepositoryType;
+import org.springframework.core.convert.converter.Converter;
+
+public class ArtifactoryRepositoryTypeConverter
+    implements Converter<String, ArtifactoryRepositoryType> {
+
+  @Override
+  public ArtifactoryRepositoryType convert(String source) {
+    return ArtifactoryRepositoryType.valueOf(source.toUpperCase());
+  }
+}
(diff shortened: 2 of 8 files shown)


igor-web/src/main/java/com/netflix/spinnaker/igor/artifactory/model/ArtifactoryRepositoryType.java
```diff
@@ -17,5 +17,6 @@
 package com.netflix.spinnaker.igor.artifactory.model;
 
 public enum ArtifactoryRepositoryType {
-  Maven,
+  MAVEN,
```
robzienert: 🤔 I believe this change is backwards-incompatible. I agree that enums should be capitalized, but if this enum has been serialized anywhere, we'll need to store the value as a string instead...

```java
public enum ArtifactoryRepositoryType {
  MAVEN("Maven"),
  HELM("HELM");

  // ...
}

Comment (anz-ecp) [reply igor-web/src/main/java/com/netflix/spinnaker/igor/artifactory/model/ArtifactoryRepositoryType.java]: Thank you  @robzienert good point , I will address that.

Comment (sirigithub) [reply igor-web/src/main/java/com/netflix/spinnaker/igor/artifactory/model/ArtifactoryRepositoryType.java]: Hi @robzienert thank you for the review.

Comment (sirigithub) [reply igor-web/src/main/java/com/netflix/spinnaker/igor/artifactory/model/ArtifactoryRepositoryType.java]: > 🤔 I believe this change is backwards-incompatible. I agree that enums should be capitalized, but if this enum has been serialized anywhere, we'll need to store the value as a string instead...
> 
> ```java
> public enum ArtifactoryRepositoryType {
>   MAVEN("Maven"),
>   HELM("HELM");
> 
>   // ...
> }
> ```

Just so I understand , so the implementation will be 
`
public enum ArtifactoryRepositoryType {
    MAVEN("Maven"),
    HELM("HELM");
    String repoTypeString;
    ArtifactoryRepositoryType(String repoTypeString)
    {
       this.repoTypeString = repoTypeString;
    }
    public String getRepoTypeString() {
        return repoTypeString;
    }
}
`
To map the search type in ArtifactorySearch , it could still be 
`public ArtifactoryRepositoryType getRepositoryType() {
    return ArtifactoryRepositoryType.valueOf(repoType.toUpperCase());
  }`
The references to ArtifactoryRepositoryType in ArtifactoryBuildMonitor and ArtifactoryItem could use the enum or be updated to use the string value instead repoType.getRepoTypeString()

Would that be accurate?

**Comment by sirigithub** [comment]
> * Fix backwards-incompatible change
> * Delete the `Task` file in the root.

@robzienert :

- Deleted Task file
- Modified ArtifactoryRepositoryType to include a String
- Updated  switch in ArtifactoryItem to use the above repo type String.
Please do let me know your thoughts.

File: igor-web/src/main/java/com/netflix/spinnaker/igor/artifactory/model/ArtifactoryItem.java [hunk 2]
```diff
@@ -32,9 +32,9 @@
   private List<ArtifactoryArtifact> artifacts;
 
   @Nullable
-  public Artifact toMatchableArtifact(ArtifactoryRepositoryType repoType, String baseUrl) {
-    switch (repoType) {
-      case Maven:
+  public Artifact toMatchableArtifact(ArtifactoryRepositoryType repositoryType, String baseUrl) {
+    switch (repositoryType.getRepoTypeString()) {
+      case "Maven":
```
robzienert: Not sure why you're converting these to a string. By doing so, you'll remove IntelliJ's (and other IDEs) ability to know if the switch statement is exhaustive or not. I'd recommend changing this back to switching on the enum type.

igor-web/src/main/java/com/netflix/spinnaker/igor/artifactory/model/ArtifactoryRepositoryType.java
```diff
@@ -13,9 +13,18 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-
 package com.netflix.spinnaker.igor.artifactory.model;
 
 public enum ArtifactoryRepositoryType {
-  Maven,
+  MAVEN("Maven"),
```
robzienert: 👍

igor-web/src/main/java/com/netflix/spinnaker/igor/artifactory/model/ArtifactorySearch.java
```diff
@@ -22,7 +22,7 @@
 @Data
 public class ArtifactorySearch {
   private String name;
-  private ArtifactoryRepositoryType repoType = ArtifactoryRepositoryType.Maven;
+  private String repoType;
```
robzienert: Keep this as the enum.

igor-web/src/main/java/com/netflix/spinnaker/igor/artifactory/model/ArtifactoryItem.java
sirigithub: Agree, reverted change.

robzienert:
This looks great! A PR came in that moved some files around, so the PR needs to be updated. Once that's done, this can get merged.

igor-web/src/main/java/com/netflix/spinnaker/igor/artifactory/model/ArtifactorySearch.java
sirigithub: Done.

igor-web/src/main/java/com/netflix/spinnaker/igor/artifactory/model/ArtifactoryRepositoryType.java
sirigithub: @robzienert , on further thought and testing , moved the artifact extension search field to the enum as well as this will remain a constant for each repo type.

Updated Halyard to exclude the field from config too.

sirigithub:
@robzienert ,updated changes as per review suggestions. On further thought and testing , moved the artifact extension search field to the enum as well as this will remain a constant for each repo type.
Updated Halyard to exclude the field from config too.

sirigithub:
> This looks great! A PR came in that moved some files around, so the PR needs to be updated. Once that's done, this can get merged.

Thank you Rob. Are you able to please look at the accompanying Halyard change too spinnaker/halyard#1472?

ryu1kn:
Rebased onto master to follow the change that moved files around.

PR #573, opened 2019-12-13
feat(managed-delivery): API to retrieve Managed Delivery manifests

luispollo:
Adds a couple of new API endpoints to list and retrieve Delivery Config manifests from source control repositories to support a new pipeline stage which pulls the manifests from source and pushes them to the `keel` API, in support of GitOps flows.

For additional context, the high-level flow for the stage would be:
1. Pipeline is triggered upon changes to SCM repo. The stage will have information about where to find the Delivery Config manifest in the repo.
2. `orca` calls `igor` to retrieve the Delivery Config manifest and publishes it to `keel`

TO-DO:
- [x] Tests
- [x] Address review feedback


igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/AbstractScmMaster.groovy
```diff
@@ -0,0 +1,29 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.igor.scm
+
+import sun.reflect.generics.reflectiveObjects.NotImplementedException
```
luispollo: Oops, that was not the right exception type at all! 😝

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy
```diff
@@ -0,0 +1,89 @@
...
+  List<String> listKeelManifests(@PathVariable String scmType,
+                                 @PathVariable String projectKey,
+                                 @PathVariable String repositorySlug,
+                                 @RequestParam(required = false, defaultValue = 'refs/heads/master') String at) {
+    String path = KEEL_MANIFESTS_BASE_PATH
+    log.info("Listing keel manifests at ${projectKey}:${repositorySlug}/${path}")
+    return getScmMaster(scmType).listDirectory(projectKey, repositorySlug, path, at)
+  }
+
+  @GetMapping(path = '/{scmType}/{projectKey}/{repositorySlug}/manifests/{manifest}')
```
luispollo: The full request path for this looks like `keel/stash/SPKR/keeldemo/manifests/cluster.yml`, which I don't love, but it's easy to test in the browser, etc... If you feel it would be better, I could change all the `@PathVariable`s to `@RequestParam`s.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy
```diff
@@ -0,0 +1,89 @@
...
+
+  @Autowired(required = false)
+  GitLabMaster gitLabMaster
+
+  @Autowired(required = false)
+  BitBucketMaster bitBucketMaster
+
+  static final String KEEL_MANIFESTS_BASE_PATH = ".netflix/spinnaker"
+
+  // Note: we expose a restricted endpoint here (as opposed to a full pass-thru to the SCM APIs) to limit the exposure
```
luispollo: This is why I didn't just expose a pass-thru to get contents from the SCM repos.

luispollo:
> We prefer that non-test backend code be written in Java or Kotlin, rather than Groovy. The following files have been added and written in Groovy:
> 
> * igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/AbstractScmMaster.groovy
> * igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy
> * igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/ScmMaster.groovy
> * igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/stash/client/model/DirectoryListingResponse.groovy
> * igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/stash/client/model/TextLinesResponse.groovy
> 
> See our server-side [commit conventions here](https://www.spinnaker.io/community/contributing/back-end-code/#choice-of-language).

Oh, dang it! 🤦‍♂

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/stash/client/StashClient.groovy
```diff
@@ -31,5 +34,23 @@ interface StashClient {
         @Path('projectKey') String projectKey,
         @Path('repositorySlug') String repositorySlug,
         @QueryMap Map queryMap)
+
+  // TODO: pagination support
```
luispollo: I don't think this is a big deal for now as I don't think we'll have that many manifests.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/stash/client/StashClient.groovy
```diff
@@ -31,5 +34,23 @@ interface StashClient {
...
+  @GET('/rest/api/1.0/projects/{projectKey}/repos/{repositorySlug}/browse/{dirPath}')
+  DirectoryListingResponse listDirectory(
+    @Path('projectKey') String projectKey,
+    @Path('repositorySlug') String repositorySlug,
+    @Path(value = 'dirPath', encode = false) String dirPath,
+    @Query('at') String at
+  )
+
+  // TODO: pagination support
```
luispollo: This might be needed if the manifests start to grow very large, especially with delivery configs, but I don't plan to add it as part of this PR.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy
```diff
@@ -0,0 +1,89 @@
...
+  // of accessing repo contents with igor's credentials
+
+  @GetMapping(path = '/{scmType}/{projectKey}/{repositorySlug}/manifests')
+  List<String> listKeelManifests(@PathVariable String scmType,
```
luispollo: Sample response (yes, it _is_ that simple):
```json
[
  "clb.yml",
  "cluster.yml",
  "image.yml",
  "sg-elb.yml",
  "sg.yml"
]
```

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy
```diff
@@ -0,0 +1,89 @@
...
+  @Autowired(required = false)
+  GitHubMaster gitHubMaster
+
+  @Autowired(required = false)
+  GitLabMaster gitLabMaster
+
+  @Autowired(required = false)
+  BitBucketMaster bitBucketMaster
+
+  static final String KEEL_MANIFESTS_BASE_PATH = ".netflix/spinnaker"
```
luispollo: Instead of allowing users to specify the manifest directory, I feel like we should enforce a convention, whatever it is.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy
```diff
@@ -0,0 +1,94 @@
...
+import org.springframework.web.bind.annotation.RequestMapping
+import org.springframework.web.bind.annotation.RequestParam
+import org.springframework.web.bind.annotation.RestController
+
+/**
+ * Exposes APIs to retrieve keel declarative manifests from source control repos.
+ */
+@RestController
+@Slf4j
+@RequestMapping("/keel")
```
emjburns: I think this is the wrong request mapping. Even though these endpoints are for keel, they should be descriptive of what they do

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy
emjburns: I think all the things on the igor side should be generic.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy
emjburns: maybe `scm/manifests`?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy
```diff
@@ -0,0 +1,94 @@
...
+  // Note: we expose a restricted endpoint here (as opposed to a full pass-thru to the SCM APIs) to limit the exposure
+  // of accessing repo contents with igor's credentials
+
+  @GetMapping(path = '/{scmType}/{projectKey}/{repositorySlug}/manifests')
+  List<String> listKeelManifests(@PathVariable String scmType,
+                                 @PathVariable String projectKey,
+                                 @PathVariable String repositorySlug,
+                                 @RequestParam(required = false, defaultValue = 'refs/heads/master') String at) {
+    String path = KEEL_MANIFESTS_BASE_PATH
+    log.info("Listing keel manifests at ${projectKey}:${repositorySlug}/${path}")
```
emjburns: should be debug level probably

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy
```diff
@@ -0,0 +1,94 @@
...
+
+  ObjectMapper yamlMapper = new ObjectMapper(new YAMLFactory()).
+    disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
+
+  static final String KEEL_MANIFESTS_BASE_PATH = ".netflix/spinnaker"
+
+  // Note: we expose a restricted endpoint here (as opposed to a full pass-thru to the SCM APIs) to limit the exposure
+  // of accessing repo contents with igor's credentials
+
+  @GetMapping(path = '/{scmType}/{projectKey}/{repositorySlug}/manifests')
```
emjburns: I'd like for `keel` to be removed from all these function names

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy
emjburns: There could be a default in the igor config if we decide it should be set on the igor side (maybe that's the best idea to limit exposure)?  but I think it should default to `.spinnaker/`

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy
```diff
@@ -0,0 +1,94 @@
...
+import org.springframework.web.bind.annotation.RestController
+
+/**
+ * Exposes APIs to retrieve keel declarative manifests from source control repos.
+ */
+@RestController
+@Slf4j
+@RequestMapping("/keel")
+class KeelManifestsController {
+  @Autowired(required = false)
```
emjburns: We like to do these now as constructor parameters wrapped in optional instead of autowired fields.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy
```diff
@@ -0,0 +1,94 @@
...
+  @GetMapping(path = '/{scmType}/{projectKey}/{repositorySlug}/manifests/{manifest}')
+  Map<String, Object> getKeelManifest(@PathVariable String scmType,
+                         @PathVariable String projectKey,
+                         @PathVariable String repositorySlug,
+
+
+                         @PathVariable String manifest,
+                         @RequestParam(required = false, defaultValue = 'refs/heads/master') String at) {
+    String path = "${KEEL_MANIFESTS_BASE_PATH}/${manifest}"
+    log.info("Retrieving keel manifest from ${projectKey}:${repositorySlug}/${path}")
```
emjburns: debug level probably

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy
```diff
@@ -0,0 +1,94 @@
...
+    String path = KEEL_MANIFESTS_BASE_PATH
+    log.info("Listing keel manifests at ${projectKey}:${repositorySlug}/${path}")
+    return getScmMaster(scmType).listDirectory(projectKey, repositorySlug, path, at)
+  }
+
+  @GetMapping(path = '/{scmType}/{projectKey}/{repositorySlug}/manifests/{manifest}')
+  Map<String, Object> getKeelManifest(@PathVariable String scmType,
+                         @PathVariable String projectKey,
+                         @PathVariable String repositorySlug,
+
```
emjburns: extra spaces

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy
gal-yardeni: +1 to change the mapping, it might be useful in other cases as well

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy
gal-yardeni: suggestion - since I'm not sure that all of these path params are mandatory, should we use the same convention like the boost apis, meaning convert them to query params and add the key? example:
https://<>?repoSlug=some-app&projectKey=~spkr

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy
gal-yardeni: See my above comment ^

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/stash/client/StashMaster.groovy
```diff
@@ -31,42 +35,38 @@ import javax.validation.Valid
...
+      return stashClient.listDirectory(projectKey, repositorySlug, path, at).toChildFilenames()
+    } catch (RetrofitError e) {
+      if (e.getKind() == RetrofitError.Kind.NETWORK) {
+        throw new NotFoundException("Could not find the server ${baseUrl}")
+      }
+      log.error(
+        "Failed to fetch file from {}/{}/{}, reason: {}",
+        projectKey, repositorySlug, path, e.message
+      )
+      throw e
```
gal-yardeni: should we throw the exception here? isn't it risky?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/stash/client/model/TextLinesResponse.groovy
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.igor.scm.stash.client.model
+
+import java.util.stream.Collectors
+
+class TextLinesResponse extends AbstractStashResponse {
+  List<Map<String, String>> lines
```
gal-yardeni: hmm maybe it's better to set a default value for `lines` and then  remove     if (lines == null) {
      return "null" ? I'm not a fan of this check :)

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy
luispollo: I like `.spinnaker`, but was following the example of `keeldemo`. Also, I thought when we saw the `newt` demo earlier today, the file with declarative configs was `.netflix/spinnaker.yml`? I don't have a strong opinion one way or another -- we just need to settle on _one_ convention. 😃

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy
luispollo: Like I said here, I struggled with naming for this URI... https://github.com/spinnaker/orca/pull/3344#discussion_r358500249

I think `manifests` is _too_ generic. What kind of manifests? Maybe `md-manifests` then?

luispollo: Looks like we settled on `delivery-configs`, @gal-yardeni and @emjburns. Thanks for the feedback!

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy
luispollo: The path parameters are indeed all mandatory, but changing them to query parameters is something I also mentioned it in another comment because it makes the URI more readable. Will make that change.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy
luispollo: Ah, OK. Thanks for the heads-up!

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy
luispollo: I'll have to convert all these to Java anyway... 🤷‍♂

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/stash/client/StashMaster.groovy
luispollo: The general behavior when calling these retrofit clients, AFAIK, is to just let exceptions bubble up to get handled upstream. The only reason we even have this `catch` block here is to convert the network error into a not-found, so for other cases we do need to either bubble up the original exception, or convert it to another exception type (which I wasn't aware of any).

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/DeliveryConfigManifestsController.java
```diff
@@ -0,0 +1,135 @@
...
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+
+/** Exposes APIs to retrieve keel declarative manifests from source control repos. */
+@RestController
+@EnableConfigurationProperties(DeliveryConfigProperties.class)
+@RequestMapping("/delivery-config/manifests")
```
luispollo: @emjburns @gal-yardeni Is this better?

igor-web/src/main/java/com/netflix/spinnaker/igor/scm/DeliveryConfigManifestsController.java
```diff
@@ -0,0 +1,161 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+
+/** Exposes APIs to retrieve Managed Delivery declarative manifests from source control repos. */
+@RestController
+@EnableConfigurationProperties(DeliveryConfigProperties.class)
+@RequestMapping("/delivery-config")
```
gal-yardeni: oh I like it much better now :)

igor-web/src/main/java/com/netflix/spinnaker/igor/scm/ScmMaster.java
```diff
@@ -0,0 +1,14 @@
+package com.netflix.spinnaker.igor.scm;
+
+import java.util.List;
+
+/** Abstracts underlying implementation details of each SCM system under a common interface. */
+public interface ScmMaster {
+  public static final String DEFAULT_GIT_REF = "refs/heads/master";
+
+  public abstract List<String> listDirectory(
```
gal-yardeni: I might be mistaken here, but if this is an interface. why this methods are abstracts?

luispollo: That was the IntelliJ converter thingie. Thanks for catching it -- will fix!

luispollo:
@erikmunson Just FYI -- I think you got plenty on your plate already and this won't be consumed by the UI (although we could potentially implement the UI in a way that lists the available manifests first, then lets the user pick), but feel free to provide feedback.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/stash/client/StashMaster.groovy
```diff
@@ -16,57 +16,46 @@
...
 /**
  * Wrapper class for a collection of Stash clients
  */
-class StashMaster {
+@Slf4j
+class StashMaster extends AbstractScmMaster {
     StashClient stashClient
     String baseUrl
 
-    @Bean
```
luispollo: I removed these `@Configuration` methods from here because they were redundant with `StashConfig`.

igor-web/src/main/java/com/netflix/spinnaker/igor/config/DeliveryConfigProperties.java
```diff
@@ -0,0 +1,26 @@
...
+
+package com.netflix.spinnaker.igor.config;
+
+import lombok.Data;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@Data
+@ConfigurationProperties
+public class DeliveryConfigProperties {
+  String manifestBasePath = ".netflix";
```
luispollo: This is configurable now, but defaults to `.netflix`.

erikmunson: nit: can we have it default to something that isn't `.netflix`? IMO our defaults should be appropriate for any company, and if there's no default that's appropriate for any company it should be a required bit of config vs. optional.

igor-web/src/main/java/com/netflix/spinnaker/igor/scm/DeliveryConfigManifestsController.java
```diff
@@ -0,0 +1,181 @@
...
+      @RequestParam final String project,
+      @RequestParam final String repository,
+      @RequestParam(required = false, defaultValue = ".") final String directory,
+      @RequestParam(required = false, defaultValue = "yml") String extension,
+      @RequestParam(required = false, defaultValue = ScmMaster.DEFAULT_GIT_REF) final String ref) {
+
+    // TODO: this started as a method to retrieve a list of keel resource manifests, but morphed
+    //  into supposedly returning only "delivery config" manifests, even though there's not really
+    //  a good way to find out if a file is a delivery config or not, even if we were to parse each
+    //  file (it doesn't have a Kubernetes-like apiVersion  and kind like other resources.
```
erikmunson: er... should it though? maybe that's a separate but valuable conversation?

erikmunson: like, maybe not a `kind`, but an `apiVersion` seems totally reasonable if we want to do breaking schema updates over time.

igor-web/src/main/java/com/netflix/spinnaker/igor/config/DeliveryConfigProperties.java
luispollo: Absolutely! I was thinking one thing and wrote another. Twice! 🤦‍♂️ It was supposed to be `.spinnaker`. Will make the change!

igor-web/src/main/java/com/netflix/spinnaker/igor/scm/DeliveryConfigManifestsController.java
luispollo: I think so. I never quite understood why the delivery config wasn’t modeled like resources. IMHO, anything that is part of an external contract with users should have an API version and type.

igor-web/src/main/java/com/netflix/spinnaker/igor/scm/DeliveryConfigManifestsController.java
```diff
@@ -0,0 +1,181 @@
...
+
+  private final DeliveryConfigProperties configProperties;
+  private final ObjectMapper jsonMapper;
+  private final ObjectMapper yamlMapper;
+  private final Optional<StashMaster> stashMaster;
+  private final Optional<GitHubMaster> gitHubMaster;
+  private final Optional<GitLabMaster> gitLabMaster;
+  private final Optional<BitBucketMaster> bitBucketMaster;
+
+  public DeliveryConfigManifestsController(
```
gal-yardeni: Comment on the controller: maybe it's better to move all the logic to a separate class, like DeliveryConfigManifestsService? That will keep the controller clean and will follow the same conventions as we have in other services.

igor-web/src/main/java/com/netflix/spinnaker/igor/scm/stash/client/model/DirectoryChild.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+package com.netflix.spinnaker.igor.scm.stash.client.model;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import lombok.Data;
+
+@Data
+@JsonIgnoreProperties(ignoreUnknown = true)
+public class DirectoryChild {
```
gal-yardeni: Can you please add comments on what is DirectoryChild and DirectoryChildren?

gal-yardeni:
Look good to me! added 2 small comments but it looks great otherwise.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/bitbucket/client/model/AbstractBitBucketResponse.groovy
```diff
@@ -17,7 +17,6 @@
 package com.netflix.spinnaker.igor.scm.bitbucket.client.model
 
 abstract class AbstractBitBucketResponse {
-  List<?> values
```
emjburns: just curious, why are you removing this?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/stash/client/model/AbstractStashResponse.groovy
```diff
@@ -20,7 +20,6 @@ abstract class AbstractStashResponse {
     int size
     int limit
     boolean isLastPage
-    List<?> values
```
emjburns: I'm also curious why you removed this

igor-web/src/main/java/com/netflix/spinnaker/igor/scm/DeliveryConfigManifestsController.java
emjburns: (1) we're no longer allowing individual resources, so this todo is confusing. There will be only one file type.
(2) no version on delivery configs is a super valid point and probably worth discussing - let's create an issue for it.

luispollo: @emjburns I know we’re not, but I thought we agreed to restrict that on the orca side, not igor’s. If we want to restrict it here and enforce the `spinnaker.yml` convention I’m cool with that, but then I’ll remove the list API and rename the other. BTW, I kinda think we’d be hijacking `spinnaker.yml` for a specific feature, so I would suggest `delivery.yml` instead, given the contents of the file. Thoughts on all of the above?

igor-web/src/main/java/com/netflix/spinnaker/igor/scm/DeliveryConfigManifestsController.java
luispollo: I didn’t think there was a logic here to warrant a new service object, plus there’s already some processing in the new `ScmMaster` methods, but if it’s a pattern I’d definitely make that change.

igor-web/src/main/java/com/netflix/spinnaker/igor/scm/stash/client/model/DirectoryChild.java
luispollo: These are just representations of the elements in the response from the stash API (which are not documented BTW). Children is a collection of “childs” basically. I can add some comments.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/KeelManifestsController.groovy
emjburns: It should be configurable in the igor config regardless. `.netflix/` will be what it is internal to netflix, but that shouldn't be the default for oss. I think the default for oss should be `.spinnaker/`

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/bitbucket/client/model/AbstractBitBucketResponse.groovy
luispollo: Because it turns out that not all stash responses have that at the root, including the ones I used for the new APIs.

luispollo: Actually, this was supposed to have been in the stash class only. I think this was IntelliJ trying to be helpful. Will fix — thanks for catching it.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/stash/client/model/AbstractStashResponse.groovy
luispollo: Answer in the bitbucket class.

igor-web/src/main/java/com/netflix/spinnaker/igor/scm/DeliveryConfigManifestsController.java
gal-yardeni: You can see gate for example. Its up to you I guess :)

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/stash/client/model/AbstractStashResponse.groovy
emjburns: I guess what I'm wondering is - is this being used elsewhere? will this break existing behavior? groovy doesn't always tell you if you remove a property that is in use.

igor-web/src/main/java/com/netflix/spinnaker/igor/scm/DeliveryConfigManifestsController.java
emjburns: I do think it's up to you Luis. I generally like to pull all logic to a separate class and keep the controllers thin. That way it's easier to write tests (and you can avoid testing the controllers unless you'd like to specifically test serialization behavior, etc). Specially, there's a lot of logic in `getManifest(` that would probably belong in a service class.

igor-web/src/main/java/com/netflix/spinnaker/igor/scm/DeliveryConfigManifestsController.java
emjburns: If we want to restrict this on the orca side the todo is still confusing. Maybe we can have an offline discussion about the design of this controller to understand how things should be named?

I'm unsure what you mean with "I kinda think we’d be hijacking spinnaker.yml for a specific feature".

luispollo: I’ll remove the to-do, but what do you think about restricting this to retrieve one the delivery config file in Igor? I’m inclined to go that route, actually. It was unfortunate timing for me that we removed support for resources exactly while I was working on this, but I rather take a bit extra time to get it right.

About `spinnaker.yml` I just thought there might be other things about spinnaker config users might end up putting in their repos eventually and by taking that file name for MD we’d prevent any future usage, is all.

igor-web/src/main/java/com/netflix/spinnaker/igor/scm/DeliveryConfigManifestsController.java
luispollo: Fair points. I’ll add the service class.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/stash/client/model/AbstractStashResponse.groovy
luispollo: It was not. I checked. There was a single subclass using that field.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/bitbucket/client/model/AbstractBitBucketResponse.groovy
luispollo: On second thought, it turns out this was already defined in the single sub-class, and like stash not all responses will have it at the root, so I'll keep this.

igor-web/src/main/java/com/netflix/spinnaker/igor/scm/DeliveryConfigManifestsController.java
luispollo: @emjburns I've decided to leave orca in charge of specifying the manifest name and kept the method to list manifests, just in case the UI needs it and/or we eventually support multiple delivery config files in one dir.

igor-web/src/main/java/com/netflix/spinnaker/igor/scm/DeliveryConfigManifestsController.java
luispollo: Done. Hope that looks better.

luispollo:
@emjburns and @gal-yardeni thanks again for your thorough review. I've refactored the code to introduce a service class as suggested, and ended up renaming classes as well to something I thought was better. Hope it'll make sense to you. Please take another look if you can.

igor-web/src/main/java/com/netflix/spinnaker/igor/scm/DeliveryConfigManifestsController.java
emjburns: sounds good!

PR #582, opened 2020-01-13
feat(core): Allow multiple custom decorators of the same type

jervi:
Each matched decorator will add another artifact to the build. This allows us to for instance migrate to new artifact types without breaking backwards compatibility. Example: We have historically used the type `docker` for docker images, but newer versions of Spinnaker expects Docker images to be of the type `docker/image`. Now we can define both decorators at the same time.


jervi:
Wait a minute before merging this. I found a little thing I want to fix. I’ll push it soon.

jervi:
Fixed in 85a4ca1

jervi:
We've run Igor with this PR merged internally in Schibsted for 9 days now. @gardleopard or maybe @ajordens can you please merge it? 🙏

igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericArtifact.java
```diff
@@ -45,4 +47,5 @@ public GenericArtifact(String type, String name, String version, String referenc
   private String type;
   private String version;
   private Map<String, String> metadata;
+  private boolean decorated;
```
jervi: This will serialize this field as well and surface it to the pipeline execution. I had originally annotated it with `@JsonIgnore` but I removed it again because I felt it wouldn't hurt to actually expose the field. I'm happy to re-add it if someone feels different.

jervi:
@spinnaker/oss-reviewers PTAL

xavileon:
I think @spinnaker/oss-approvers are the ones with rights to merge?

igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericArtifact.java
jervi: Actually, I found a use for the flag (in https://github.com/spinnaker/orca/pull/3422), so now I'm not so happy to ignore it anymore 😎

PR #591, opened 2020-01-21
chore(dependencies): Spring Boot 2.2.4 upgrade

pdelagrave:


igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildParser.java
```diff
@@ -48,6 +52,24 @@
     return parse(inputString, destinationClass);
   }
 
+  @SuppressWarnings("unchecked")
+  public final Map serializeBuild(Build inputBuild) {
+    // com.google.api.services.cloudbuild.v1.model.StorageSource.generation is of type Long
```
cfieber: @plumpy think you could find the right team to let them know about this?

Vity01:
Just note there is already Spring Boot 2.2.4...

pdelagrave:
> Just note there is already Spring Boot 2.2.4...

As yesterday we'll be using 2.2.4: https://github.com/spinnaker/kork/pull/482
The PR titles just haven't been updated.

pdelagrave:
flaky test but otherwise ready

igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildParser.java
pdelagrave: Still an issue but moved the fix in test code as this only impacts our tests so far.
https://github.com/spinnaker/igor/blob/c467ba2ad813c3b4abf4e0ab965ab020edbdeb86/igor-web/src/test/groovy/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildTestSerializationHelper.groovy

PR #593, opened 2020-01-21
refactor(gcb): Reduce visibility of GCB packages

ezimanyi:
* refactor(gcb): Move all GCB files into the same package 

  The GoolgeCloudBuildArtifact and GoolgeCloudBuildConfig classes were in different packages; by moving these to the same package as the other GCB classes we'll be able to reduce the visibility of many GCB classes to package-private.

* refactor(gcb): Make packages package-private and final where possible 

  The only external contract the com.netflix.spinnaker.igor.gcb package intends to provide is the GoogleCloudBuildController component; everything else is an implementation detail.

  To better communicate and enforce this:
  * Make the controller (and all mapped methods) public; some of these were package-private before which works because Spring uses reflection to call controllers, but we should document them accurately as being called outside of the package.
  * Make all other classes (and methods) that were public package-private

  Finally (no pun intended), none of the classes in this package were designed for inheritance so mark them final wherever possible. The cases where it is not possible are:
  * Classes that are mocked in tests as a final class cannot be mocked; in this case, we'll rely on the fact that the class is package-private so at least can't be subclassed outside the package.
  * Spring controller and configuration classes aren't allowed to be final

* refactor(gcb): Add nullity annotations 

  Add default Nonnull to all methods, parameters and fields across the package, and annotate the few exceptions with Nullable.

  This does surface a few places where we could consider replacing a null with either an Optional or an appropriate null object, but this PR does not tackle any of those refactors.

* refactor(gcb): A few minor refactors 

  * Replace statement lambda with expression lambda
  * Make GoogleCloudBuildExecutor nicer to clients by adding type bounds.
  * Clean up the null checks of build status by using a null-safe valueOf that returns UNKNOWN. (This keeps the existing behavior as UNKNOWN status has the lowest priority, so null status continues to have the lowest priority).
  * Use try-with-resources on JsonGenerator so that it always gets closed before we use the underlying writer, avoiding the need to explicitly flush.

* refactor(gcb): Make GCB account repository immutable 

  As we never add/remove accounts after startup, the repository should be immutable. Maybe at some point if we support dynamic accounts we should change this (but at that point we'd need to think about thread safety anyway).

  This commit also makes a few changes to the tests on that class; they had to be updated anyway to use the new Builder class, and while I was there I improved the assertions to make them more clear.

* refactor(gcb): Update GCB controller to return immutable lists 

  Return immutable lists from the controller, and convert dependent code to also use immutable lists.

### DIFF
--- a/igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildParser.java
+++ b/igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildParser.java
@@ -20,6 +20,7 @@ import com.google.api.client.json.JsonGenerator;
 import com.google.api.client.json.jackson2.JacksonFactory;
 import java.io.IOException;
 import java.io.StringWriter;
+import java.io.UncheckedIOException;
 import java.io.Writer;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.stereotype.Component;
@@ -39,7 +40,7 @@ final class GoogleCloudBuildParser {
     try {
       return jacksonFactory.createJsonParser(input).parse(destinationClass);
     } catch (IOException e) {
-      throw new RuntimeException(e);
+      throw new UncheckedIOException(e);
     }
   }
 
@@ -53,7 +54,7 @@ final class GoogleCloudBuildParser {
     try (JsonGenerator generator = jacksonFactory.createJsonGenerator(writer)) {
       generator.serialize(input);
     } catch (IOException e) {
-      throw new RuntimeException(e);
+      throw new UncheckedIOException(e);
     }
     return writer.toString();
   }
--- a/igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildParser.java
+++ b/igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildParser.java
@@ -32,10 +32,10 @@ import org.springframework.stereotype.Component;
  */
 @Component
 @ConditionalOnProperty("gcb.enabled")
-public class GoogleCloudBuildParser {
+final class GoogleCloudBuildParser {
   private final JacksonFactory jacksonFactory = JacksonFactory.getDefaultInstance();
 
-  public final <T> T parse(String input, Class<T> destinationClass) {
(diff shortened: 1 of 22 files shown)


ezimanyi:
This is a collection of things I've learned since I wrote the GCB implementation almost a year ago; I've had a branch where I've been accumulating these changes and it was time to clean it up and submit.

igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/NonnullByDefault.java
```diff
@@ -0,0 +1,32 @@
...
+import javax.annotation.meta.TypeQualifierDefault;
+
+/**
+ * Defaults all fields, methods, and method parameters to @Nonnull by default. TODO(ezimanyi): Move
+ * this to kork-annotations as it is likely useful outside of igor
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Nonnull
+@TypeQualifierDefault({ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER})
+public @interface NonnullByDefault {}
```
plumpy: Nice, way better than `@ParametersAreNonnullByDefault`!

igor-web/src/main/java/com/netflix/spinnaker/igor/gcb/GoogleCloudBuildParser.java
```diff
@@ -49,14 +49,12 @@
...
-    try {
-      Writer writer = new StringWriter();
-      JsonGenerator generator = jacksonFactory.createJsonGenerator(writer);
+    Writer writer = new StringWriter();
+    try (JsonGenerator generator = jacksonFactory.createJsonGenerator(writer)) {
       generator.serialize(input);
-      generator.flush();
-      return writer.toString();
     } catch (IOException e) {
       throw new RuntimeException(e);
```
plumpy: idea for future cleanup: `throw new UncheckedIOException(e)` (a new class in Java 8)

But I see this pattern is used all over this package, and it's of questionable value, so whatever

ezimanyi: Nice, I didn't realize that existed!  I replaced all of the `RuntimeException`s that were wrapping `IOException`s with `UncheckedIOException`s.

PR #594, opened 2020-01-22
refactor(*): Pull jenkins out into separate module; create new core abstractions

robzienert:
I did this over the holiday break. It's a continuation of my desire to see Igor be a respectable service. Of all its warts, the biggest issue with it is that every CI provider needs to walk and talk as if it is Jenkins. It's understandable, as when it was originally written it was only for Jenkins.

This PR does a few things:

1. Splits the Jenkins codebase out into a `igor-monitor-jenkins` module.
1. Creates some loose "Step 1" abstractions around core models. It's still not great, but it's better.
1. Converts all Groovy code in `main` source set to Java.

I haven't had the chance to open this up again, but I wanted to put it out there for anyone that's particularly interested in reviewing large pull requests. I tried to keep the changeset as minimal as possible, because I don't think I can correctly make the right abstractions until all providers are decoupled from each other.

Side note: I see that IntelliJ "helpfully" rewrote a bunch of license headers. They will be fixed.


igor-monitor-jenkins/src/main/java/com/netflix/spinnaker/igor/jenkins/client/model/Build.java
```diff
@@ -0,0 +1,184 @@
...
+public class Build {
+  public GenericBuild genericBuild(final String jobName) {
+    GenericBuild.GenericBuildBuilder builder =
+        GenericBuild.builder()
+            .building(building)
+            .number(number)
+            .duration(duration.intValue())
+            // TODO(rz): Groovyism. What does Groovy do when `result` is null and you cast it to an
+            // enum? WHO KNOWS, but
+            //  all of the igor tests depend on _something_ being set from null.
```
emjburns: 😆

Vity01:
Is there any chance this PR will be merged into master?

robzienert:
Yes, it'll get merged into master, but I first need to finish it. You could consider this work "20% time", except that the 20% time comes out of my evenings / nights, so I can't give you an estimate of when I'll have it done.

igor-core/src/main/java/com/netflix/spinnaker/igor/history/model/Event.java
```diff
@@ -0,0 +1,4 @@
+package com.netflix.spinnaker.igor.history.model;
+
+/** TODO(rz): Document. */
```
emjburns: Would probably be good to document these top level interfaces / how they will be used before some of the more specific things.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
```diff
@@ -146,28 +136,11 @@ class BuildController {
...
-            } catch (RetrofitError e) {
-                if (e.response?.status != NOT_FOUND.value()) {
-                    throw e
-                }
-            }
+        if (buildService instanceof BuildQueueOperations) {
+          buildService.stopQueuedBuild(jobName, queuedBuild, buildNumber)
         }
 
+        // TODO(rz): lol, for real?
```
emjburns: 😆

PR #595, opened 2020-01-22
feat(codebuild): Add support for triggering and monitoring AWS CodeBu…

Kaixiang-AWS:
…ild builds

Add functionalities to start a build in AWS CodeBuild and monitor the build status. AWS CodeBuild is different from the existing build services (Jenkins, Travis) therefore cannot use the existing architecture. It's more like Google Cloud Build which implements its own authentication strategy.

For CodeBuild account in igor, it uses AWS access key id and secret to authenticate. I don't know if there's way to talk to clouddriver to fetch aws cloud provider credentials. If yes, I can update my implementation later to fetch credentials from clouddriver.

Related Issue: spinnaker/spinnaker#5292

### DIFF
--- a/igor-web/src/main/java/com/netflix/spinnaker/igor/codebuild/AwsCodeBuildAccount.java
+++ b/igor-web/src/main/java/com/netflix/spinnaker/igor/codebuild/AwsCodeBuildAccount.java
@@ -16,26 +16,36 @@
 
 package com.netflix.spinnaker.igor.codebuild;
 
-import com.amazonaws.auth.AWSStaticCredentialsProvider;
-import com.amazonaws.auth.BasicAWSCredentials;
+import com.amazonaws.auth.STSAssumeRoleSessionCredentialsProvider;
 import com.amazonaws.services.codebuild.AWSCodeBuildClient;
 import com.amazonaws.services.codebuild.AWSCodeBuildClientBuilder;
 import com.amazonaws.services.codebuild.model.BatchGetBuildsRequest;
 import com.amazonaws.services.codebuild.model.Build;
 import com.amazonaws.services.codebuild.model.StartBuildRequest;
+import com.amazonaws.services.securitytoken.AWSSecurityTokenServiceClient;
+import java.util.Objects;
 import lombok.RequiredArgsConstructor;
+import org.springframework.beans.factory.annotation.Autowired;
 
 /** Generates authenticated requests to AWS CodeBuild API for a single configured account */
 @RequiredArgsConstructor
 public class AwsCodeBuildAccount {
   private final AWSCodeBuildClient client;
 
-  public AwsCodeBuildAccount(String accessKeyId, String secretAccessKey, String region) {
-    BasicAWSCredentials awsCreds = new BasicAWSCredentials(accessKeyId, secretAccessKey);
+  @Autowired AWSSecurityTokenServiceClient stsClient;
+
+  public AwsCodeBuildAccount(String accountId, String assumeRole, String region) {
+    STSAssumeRoleSessionCredentialsProvider credentialsProvider =
+        new STSAssumeRoleSessionCredentialsProvider.Builder(
+                getRoleArn(accountId, assumeRole), "spinnaker-session")
+            .withStsClient(stsClient)
+            .build();
+
+    // TODO: Add client-side rate limiting to avoid getting throttled if necessary
     this.client =
         (AWSCodeBuildClient)
             AWSCodeBuildClientBuilder.standard()
-                .withCredentials(new AWSStaticCredentialsProvider(awsCreds))
+                .withCredentials(credentialsProvider)
                 .withRequestHandlers(new AwsCodeBuildRequestHandler())
                 .withRegion(region)
                 .build();
@@ -47,5 +57,20 @@ public class AwsCodeBuildAccount {
 
   public Build getBuild(String buildId) {
     return client.batchGetBuilds(new BatchGetBuildsRequest().withIds(buildId)).getBuilds().get(0);
+  }
+
+  private String getRoleArn(String accountId, String assumeRole) {
+    String assumeRoleValue = Objects.requireNonNull(assumeRole, "assumeRole");
+    if (!assumeRoleValue.startsWith("arn:")) {
+      /**
+       * GovCloud and China regions need to have the full arn passed because of differing formats
+       * Govcloud: arn:aws-us-gov:iam China: arn:aws-cn:iam
+       */
+      assumeRoleValue =
+          String.format(
+              "arn:aws:iam::%s:%s",
+              Objects.requireNonNull(accountId, "accountId"), assumeRoleValue);
+    }
+    return assumeRoleValue;
   }
 }
--- a/igor-web/src/main/java/com/netflix/spinnaker/igor/codebuild/AwsCodeBuildAccount.java
+++ b/igor-web/src/main/java/com/netflix/spinnaker/igor/codebuild/AwsCodeBuildAccount.java
@@ -1,0 +1,51 @@
+/*
+ * Copyright 2020 Amazon.com, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.igor.codebuild;
+
+import com.amazonaws.auth.AWSStaticCredentialsProvider;
+import com.amazonaws.auth.BasicAWSCredentials;
+import com.amazonaws.services.codebuild.AWSCodeBuildClient;
+import com.amazonaws.services.codebuild.AWSCodeBuildClientBuilder;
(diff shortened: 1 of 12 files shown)


igor-web/src/main/java/com/netflix/spinnaker/igor/codebuild/AwsCodeBuildAccount.java
```diff
@@ -0,0 +1,51 @@
...
+@RequiredArgsConstructor
+public class AwsCodeBuildAccount {
+  private final AWSCodeBuildClient client;
+
+  public AwsCodeBuildAccount(String accessKeyId, String secretAccessKey, String region) {
+    BasicAWSCredentials awsCreds = new BasicAWSCredentials(accessKeyId, secretAccessKey);
+    this.client =
+        (AWSCodeBuildClient)
+            AWSCodeBuildClientBuilder.standard()
+                .withCredentials(new AWSStaticCredentialsProvider(awsCreds))
```
clareliguori: Ideally you don't use static credentials, you use a role so that you don't have to store static creds in your hal config.  With clouddriver, the 'aws' provider account is configured with a role to assume, and the environment where clouddriver runs is configured with credentials that the SDK default creds provider can pick up to assume that role (like instance creds, ECS creds, env variable creds, etc) using the STSAssumeRoleSessionCredentialsProvider.

So my hal config looks something like this:
```
      accounts:
      - name: my-aws-devel-acct
        requiredGroupMembership: []
        providerVersion: V1
        accountId: '111111111111'
        regions:
        - name: eu-central-1
        assumeRole: role/SpinnakerManaged
```

And then clouddriver just uses the role session provider:
https://github.com/spinnaker/clouddriver/tree/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security

igor-web/src/main/java/com/netflix/spinnaker/igor/codebuild/AwsCodeBuildMapper.java
```diff
@@ -0,0 +1,29 @@
...
+
+import com.amazonaws.services.codebuild.model.StartBuildRequest;
+import java.util.Map;
+
+public class AwsCodeBuildMapper {
+  private static final String PROJECT_NAME = "projectName";
+
+  public static StartBuildRequest toStartBuildRequest(Map<String, Object> requestBody) {
+    String projectName = (String) requestBody.get(PROJECT_NAME);
+    return new StartBuildRequest().withProjectName(projectName);
```
clareliguori: When this is a stage in a pipeline, is it common to get any other attributes like the commit ID that needs to be built?  I'm surprised not to see sourceVersion as an input

igor-web/src/main/java/com/netflix/spinnaker/igor/codebuild/AwsCodeBuildRequestHandler.java
```diff
@@ -0,0 +1,39 @@
...
+import com.amazonaws.handlers.RequestHandler2;
+import com.amazonaws.services.codebuild.model.AWSCodeBuildException;
+import com.netflix.spinnaker.igor.exceptions.BuildJobError;
+import lombok.extern.slf4j.Slf4j;
+
+@Slf4j
+public class AwsCodeBuildRequestHandler extends RequestHandler2 {
+  @Override
+  public void afterError(Request<?> request, Response<?> response, Exception e) {
+    if (e instanceof AWSCodeBuildException) {
```
clareliguori: Would this include ThrottlingException?  Should API rate throttling be a build job error or a runtime exception?

igor-web/src/main/java/com/netflix/spinnaker/igor/config/AwsCodeBuildProperties.java
```diff
@@ -0,0 +1,35 @@
...
+@ConfigurationProperties(prefix = "codebuild")
+public class AwsCodeBuildProperties {
+  private List<Account> accounts;
+
+  @Data
+  public static class Account {
+    private String name;
+    private String region;
+    private String accessKeyId;
+    private String secretAccessKey;
```
clareliguori: Ideally this is a role

igor-web/src/main/java/com/netflix/spinnaker/igor/codebuild/AwsCodeBuildAccount.java
```diff
@@ -0,0 +1,51 @@
...
+/** Generates authenticated requests to AWS CodeBuild API for a single configured account */
+@RequiredArgsConstructor
+public class AwsCodeBuildAccount {
+  private final AWSCodeBuildClient client;
+
+  public AwsCodeBuildAccount(String accessKeyId, String secretAccessKey, String region) {
+    BasicAWSCredentials awsCreds = new BasicAWSCredentials(accessKeyId, secretAccessKey);
+    this.client =
+        (AWSCodeBuildClient)
+            AWSCodeBuildClientBuilder.standard()
```
clareliguori: I'm not sure if this is needed here, but clouddriver has a framework where the client can rate-limit itself to avoid getting throttled.  This is more because clouddriver tends to hammer Describe APIs with polling, which you'll be avoiding.  But it might be interesting to look at (or mark as a TODO if people run into throttling later):
https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/sdkclient/RateLimiterSupplier.java

https://github.com/spinnaker/clouddriver/blob/master/clouddriver-aws/src/main/groovy/com/netflix/spinnaker/clouddriver/aws/security/sdkclient/RateLimitingRequestHandler.java

https://docs.armory.io/spinnaker-install-admin-guides/rate-limit/

igor-web/src/main/java/com/netflix/spinnaker/igor/codebuild/AwsCodeBuildAccount.java
dipper-github-preprod: The ideal scenario would be specifying aws account name in codebuild account configuration and then all codebuild operations use the credentials from that aws account associated. However, I couldn't find a way to let igor talk to clouddriver to fetch the credentials. Another way is re-implementing the assume role logic in igor. That will introduce duplicate code in two packages. I can do that if that's really necessary.

Kaixiang-AWS: The ideal scenario would be specifying aws account name in codebuild account configuration and then all codebuild operations use the credentials from that aws account associated. However, I couldn't find a way to let igor talk to clouddriver to fetch the credentials. Another way is re-implementing the assume role logic in igor. That will introduce duplicate code in two packages. I can do that if that's really necessary.

igor-web/src/main/java/com/netflix/spinnaker/igor/codebuild/AwsCodeBuildMapper.java
Kaixiang-AWS: I will add more advanced use case such as source version and environment variables in later commits. This commit is more focusing on successfully trigger a build from Spinnaker and monitor build status.

igor-web/src/main/java/com/netflix/spinnaker/igor/codebuild/AwsCodeBuildRequestHandler.java
Kaixiang-AWS: ThrottlingException does not inherit from AWSCodeBuildException so it will be a RuntimeException

igor-web/src/main/java/com/netflix/spinnaker/igor/codebuild/AwsCodeBuildAccount.java
Kaixiang-AWS: I will add a TODO here.

igor-web/src/main/java/com/netflix/spinnaker/igor/codebuild/AwsCodeBuildAccount.java
```diff
@@ -0,0 +1,51 @@
...
+                .withCredentials(new AWSStaticCredentialsProvider(awsCreds))
+                .withRequestHandlers(new AwsCodeBuildRequestHandler())
```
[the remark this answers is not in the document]
Kaixiang-AWS: Nice catch. I will inject user agent info into the client.

igor-web/src/main/java/com/netflix/spinnaker/igor/codebuild/AwsCodeBuildAccount.java
clareliguori: Yeah agree that would be ideal, but I do think you will need to duplicate the logic unfortunately, though it can probably be simpler in Igor than the way it's implemented in clouddriver

igor-web/src/main/java/com/netflix/spinnaker/igor/codebuild/AwsCodeBuildRequestHandler.java
clareliguori: I'm not sure what the implications of picking build job error vs runtime exception are for how Spinnaker behaves when these are returned; I'm not super familiar with Igor.  Why should ThrottlingException be runtime error but AccountLimitExceededException be a build error?

Kaixiang-AWS: BuildJobError is returned as 400 while RuntimeException will become an InternalError and return 500. I think throttling exception should probably return 400 instead of 500. I will categorize it as BuildJobError as well.

clareliguori: You could also look at the exception error type:
https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/AmazonServiceException.ErrorType.html

PR #624, opened 2020-02-12
feat(codebuild): Add getArtifacts API for AWS CodeBuild

Kaixiang-AWS:
This API will be called by orca to fetch S3 artifacts produced by CodeBuild
so that these artifacts could be binded and consumed by downstream stages.

### DIFF
--- a/igor-web/src/main/java/com/netflix/spinnaker/igor/codebuild/AwsCodeBuildAccount.java
+++ b/igor-web/src/main/java/com/netflix/spinnaker/igor/codebuild/AwsCodeBuildAccount.java
@@ -21,9 +21,13 @@ import com.amazonaws.services.codebuild.AWSCodeBuildClient;
 import com.amazonaws.services.codebuild.AWSCodeBuildClientBuilder;
 import com.amazonaws.services.codebuild.model.BatchGetBuildsRequest;
 import com.amazonaws.services.codebuild.model.Build;
+import com.amazonaws.services.codebuild.model.BuildArtifacts;
 import com.amazonaws.services.codebuild.model.StartBuildRequest;
 import com.amazonaws.services.codebuild.model.StopBuildRequest;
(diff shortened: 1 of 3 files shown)


igor-web/src/main/java/com/netflix/spinnaker/igor/codebuild/AwsCodeBuildAccount.java
```diff
@@ -78,4 +87,28 @@ private String getRoleArn(String accountId, String assumeRole) {
...
+    if (secondaryArtifacts != null) {
+      secondaryArtifacts.forEach(
+          artifacts -> artifactsList.add(getS3Artifact(artifacts.getLocation())));
+    }
+    return artifactsList;
+  }
+
+  private Artifact getS3Artifact(String s3Arn) {
+    String reference = getS3ArtifactReference(s3Arn);
+    return Artifact.builder().type("s3/object").reference(reference).name(reference).build();
```
clareliguori: I could be wrong, but I think typically the artifact name is fairly static so that later stages in the pipeline can refer to that name in their definitions, and the version field is used to distinguish between artifacts from different builds.  Can the CodeBuild's namespacing behavior be used here to have a static artifact name and changing version field?

Kaixiang-AWS: Based on the [documentation](https://www.spinnaker.io/reference/artifacts-with-artifactsrewrite/types/s3-object/) `name` should be the full path to the artifact file, beginning with s3://, which is the same as `reference`. `version` is not a valid field for this type of object.

Kaixiang-AWS: Later stages could refer to this artifact by its **display name** defined in UI, which is different from the name here.

PR #626, opened 2020-02-13
fix(jenkins): Fix build event details serialization

robzienert:
I'm not sure why, but when converting the Jenkins code to Java, the `getDetails()` default method on `BuildEvent` was not being invoked by the serde process posting events into Echo, which caused builds to be missed. Converting `BuildEvent` to an abstract class and forcing the details method to exist fixed the issue.


igor-monitor-jenkins/src/main/java/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.java
```diff
@@ -279,16 +283,52 @@ private void postEvent(final Project project, final String master) {
...
+                  }
+                  log.error(
+                      "Failed to post jenkins build event to Echo ({}:{}): {}",
+                      master,
+                      project.getName(),
+                      rawResponse,
+                      e);
+                  throw e;
+                }
+              });
```
robzienert: The exception being thrown from Retrofit was totally unhelpful, so I wrote some more error handling logic to figure out what was going on. Figure why not leave it in.

igor-monitor-jenkins/src/main/java/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.java
```diff
@@ -279,16 +283,52 @@ private void postEvent(final Project project, final String master) {
...
           // TODO(rz): Add allowAnonymous(Runnable)
           return null;
         });
   }
 
+  private String getResponseBody(RetrofitError e) throws IOException {
+    Response re = e.getResponse();
+    if (re != null && re.getBody() != null) {
+      ByteSource byteSource =
```
luispollo: Couldn't you just do `e.getBodyAs(String.class)` here?

PR #631, opened 2020-02-18
chore(buildtool): add an empty .gcloudignore file

plumpy:
By default, any files in .gitignore will not be sent to Google Cloud
Build. The line '*/build' is (incorrectly) causing it to not upload the
`igor-core/src/main/java/com/netflix/spinnaker/igor/build` directory,
which makes the build fail.


plumpy:
@spinnakerbot cherry-pick 1.18
@spinnakerbot cherry-pick 1.17

PR #633, opened 2020-02-19
chore(container builds): exclude large cache directories from GCB upload

plumpy:


plumpy:
@spinnakerbot cherry-pick 1.18
@spinnakerbot cherry-pick 1.17

PR #641, opened 2020-02-25
feat(ciView): creating the basis for ci view within Spinnaker

gal-yardeni:
This is the first extension point going towards CI view within Spinnaker.

Added a new ci controller that will be involved in the future, and a dummy service which is the extension point for the implementation. 
Any service that will implement the contract in`CiBuildService` will be a valid backend service to support the CI view component, and potentially will be pares by deck (UI) and will be shown inside of spinnaker.


igor-web/src/main/java/com/netflix/spinnaker/igor/ci/CiBuildService.java
```diff
@@ -0,0 +1,34 @@
...
+package com.netflix.spinnaker.igor.ci;
+
+import com.netflix.spinnaker.igor.build.model.GenericBuild;
+import java.util.List;
+
+public interface CiBuildService {
+
+  /**
+   * Get the builds given project key, repo slug, and status. By implementing this interface, deck
+   * will (eventually) be able to parse it and to show the details of the result in the UI.
```
emjburns: I'd change the second sentence of this comment to be something like 

> The data returned from a CiBuildService powers the CI view in the UI.

igor-web/src/main/java/com/netflix/spinnaker/igor/ci/NoopCiService.java
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.igor.ci;
+
+import com.netflix.spinnaker.igor.build.model.GenericBuild;
+import java.util.Collections;
+import java.util.List;
+
+public class NoopCiService implements CiBuildService {
```
emjburns: I don't think you need this now that we have the interface. I think that having the interface + an optional CI service will make everything compile and start up in the right way.

emjburns: The interface provides all the same information, right?

igor-web/src/main/java/com/netflix/spinnaker/igor/config/CiConfig.java
```diff
@@ -0,0 +1,34 @@
...
+
+package com.netflix.spinnaker.igor.config;
+
+import com.netflix.spinnaker.igor.ci.CiBuildService;
+import com.netflix.spinnaker.igor.ci.NoopCiService;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Bean;
+
+@ConditionalOnProperty("services.ci.enabled")
```
emjburns: similarly, i think you can delete this class in igor since it's not doing anything

emjburns: My understanding is that if you have `services.ci.enabled`, but you're not also supplying a `CiBuildService`, then when you try to use the controller it'll throw the exception.

Because we have the `Optional`,  I don't think we need the conditional on missing bean here.

robzienert:
Echo everything Emily said. Most of this can be deleted.

If by "extension point" you expect this to be plugin-compatible, there's more than needs to be done, but that can be handled separately.

igor-web/src/main/java/com/netflix/spinnaker/igor/ci/NoopCiService.java
gal-yardeni: Yep, ok make sense!

gal-yardeni:
Related gate PR: https://github.com/spinnaker/gate/pull/1078

igor-web/src/main/java/com/netflix/spinnaker/igor/ci/CiBuildService.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.netflix.spinnaker.igor.ci;
+
+import com.netflix.spinnaker.igor.build.model.GenericBuild;
+import java.util.List;
+
+public interface CiBuildService {
```
robzienert: Docs on interfaces.

PR #643, opened 2020-02-27
feat(codebuild): Add support for static credentials

Kaixiang-AWS:
After this change, user should be able to specify static credential for codebuild
and all codebuild accounts will use this credential to assume role. If assume role
is not present, the static credential itself will be used to call CodeBuild APIs.


igor-web/src/main/java/com/netflix/spinnaker/igor/config/AwsCodeBuildConfig.java
```diff
@@ -16,40 +16,91 @@
...
-            .withCredentials(DefaultAWSCredentialsProviderChain.getInstance())
+            .withCredentials(awsCredentialsProvider)
             .build();
   }
+
+  @Bean("awsCredentialsProvider")
+  AWSCredentialsProvider awsCredentialsProvider(AwsCodeBuildProperties awsCodeBuildProperties) {
+    AWSCredentialsProvider credentialsProvider = DefaultAWSCredentialsProviderChain.getInstance();
+    if (awsCodeBuildProperties.getAccessKeyId() != null
+        && !awsCodeBuildProperties.getAccessKeyId().equals("")
```
clareliguori: Nit: use isEmpty

PR #664, opened 2020-03-14
fix(artifactory/helm): Fix helm artifact type

rhysm:
This change fixes https://github.com/spinnaker/spinnaker/issues/5554. Using the exact chart name as the artifact name and the artifact type `helm/chart` allows the artifact to match the expected artifact of the Artifactory trigger.

- Change the helm artifact type to match what it is known as in other parts of Spinnaker
- Update link to be a direct download link rather than a link to the web view
- Refactor to use ArtifactoryQueryBuilder for AQL query
- Retrieve the chart name and version from the Artifactory properties


gal-yardeni:
@claymccoy can you please take a look at this PR? thanks!

PR #670, opened 2020-03-19
fix(travis): Limit the number of returned builds

jervi:
If you have builds that creates very large logs, the number of builds returned can cause issues or timeouts in the Travis API. Limiting the number of builds seems to mitigate the issue.

Drawbacks: This will limit the number of builds you'll see in the build drop down of a manual execution to 10 (but it won't affect the build monitor at all as it uses another endpoint).

### DIFF
--- a/igor-web/src/main/java/com/netflix/spinnaker/igor/travis/service/TravisService.java
+++ b/igor-web/src/main/java/com/netflix/spinnaker/igor/travis/service/TravisService.java
@@ -72,13 +72,13 @@ import retrofit.RetrofitError;
 public class TravisService implements BuildOperations, BuildProperties {
 
   static final int TRAVIS_JOB_RESULT_LIMIT = 100;
-  static final int TRAVIS_BUILD_RESULT_LIMIT = 10;
 
   private final Logger log = LoggerFactory.getLogger(getClass());
   private final String baseUrl;
   private final String groupKey;
   private final GithubAuth gitHubAuth;
   private final int numberOfJobs;
+  private final int buildResultLimit;
   private final TravisClient travisClient;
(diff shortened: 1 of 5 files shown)


igor-web/src/main/java/com/netflix/spinnaker/igor/travis/service/TravisService.java
```diff
@@ -71,7 +71,8 @@
 
 public class TravisService implements BuildOperations, BuildProperties {
 
-  static final int TRAVIS_BUILD_RESULT_LIMIT = 100;
+  static final int TRAVIS_JOB_RESULT_LIMIT = 100;
+  static final int TRAVIS_BUILD_RESULT_LIMIT = 10;
```
jervi: So I've renamed this variable everywhere _except_ where we want to limit the number of builds returned. So the code that is actually affected isn't really showing up in the changes below.

gal-yardeni: will it make more sense to have it configured in igor.yml instead of hard coded?

jervi: The second one could maybe make sense to put in a config file. The first one is the max number of results per page, so that is ok to hard code.

gal-yardeni: Yep, I think leaving the first one here it's fine, but if you can move the other param to be configurable it would be great.

jervi:
@gal-yardeni I kind of forgot about this over Easter, but I've now finally added a configurable setting for the build limit, PTAL :)

PR #673, opened 2020-03-20
fix(gcb): Return mutable lists from methods annoated with PostFilter

ezimanyi:
Fixes spinnaker/spinnaker#5593

* test(gcb): Update tests to run postFilter logic

  The GCB integration tests currently don't run the postFilter logic that is added as an annotation on controller methods. I believe this is because we've been to selective in exactly the beans that should be present in the test, and are not pulling in the required Spring Security beans.

  In order to fix this, add the same @ComponentScan that we have on Main.class to the test so it pulls in the same beans.  (We can then also remove the specific controller bean we'd been pulling in, but will leave all the configuration beans in place.)

  This causes the listAccountTest to fail because the listAccounts function is currently broken; the next commit will fix the function and the test.

* fix(gcb): Return mutable lists from methods annoated with PostFilter

  Controller methods annotated with PostFilter cannot return immutable lists because the filtering is done in-place on the returned array; update all controller methods in GoogleCloudBuildController to return a mutable list.

  As we would like the lower levels of the stack to deal in immutable collections as much as possible, just create a mutable collection at the last step before returning from the controller method.


ezimanyi:
Will look into that failing test, this is passing locally but not in Github Actions.

ezimanyi:
@spinnakerbot cherry-pick 1.19

PR #715, opened 2020-04-27
feat(plugins): igor supports plugins

link108:
This PR makes Igor pluggable, similar to https://github.com/spinnaker/clouddriver/pull/4383

### DIFF
--- a/igor-core/igor-core.gradle
+++ b/igor-core/igor-core.gradle
@@ -1,6 +1,7 @@
 dependencies {
   implementation "com.netflix.spinnaker.kork:kork-artifacts"
   implementation "com.netflix.spinnaker.kork:kork-core"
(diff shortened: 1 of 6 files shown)


igor-core/igor-core.gradle
```diff
@@ -1,6 +1,7 @@
 dependencies {
   implementation "com.netflix.spinnaker.kork:kork-artifacts"
   implementation "com.netflix.spinnaker.kork:kork-core"
+  api "com.netflix.spinnaker.kork:kork-plugins"
```
danielpeach, line 4: I think this should be `implementation` and only under `igor-web` until we write an `ExtensionPoint`.

claymccoy: That makes sense. I think we were just being consistent with how it was done in the other services.

PR #723, opened 2020-05-05
feat(monitor): Add plugins monitor

robzienert:
First step towards triggering pipelines off of new plugins. The initial implementation only supports looking at front50 as a plugin repository, but the storage schema supports future enhancements for PF4J's more general UpdateRepository format as well.

### DIFF
--- a/igor-monitor-plugins/src/main/java/com/netflix/spinnaker/igor/plugins/PluginsBuildMonitor.java
+++ b/igor-monitor-plugins/src/main/java/com/netflix/spinnaker/igor/plugins/PluginsBuildMonitor.java
@@ -1,0 +1,113 @@
+/*
+ * Copyright 2020 Netflix, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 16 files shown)


igor-monitor-plugins/src/main/java/com/netflix/spinnaker/igor/plugins/PluginsBuildMonitor.java
```diff
@@ -0,0 +1,111 @@
...
+  private final Optional<EchoService> echoService;
+
+  public PluginsBuildMonitor(
+      IgorConfigurationProperties igorProperties,
+      Registry registry,
+      Optional<DiscoveryClient> discoveryClient,
+      Optional<LockService> lockService,
+      PluginReleaseService pluginInfoService,
+      PluginCache cache,
+      Optional<EchoService> echoService) {
```
jonsie, line 50: Curious - why is EchoService is optional?

robzienert: For deployments where you don't want to send notifications - e.g. passive regional clusters.

igor-monitor-plugins/src/main/java/com/netflix/spinnaker/igor/plugins/front50/PluginInfo.java
```diff
@@ -0,0 +1,44 @@
...
+
+  final String id;
+  final List<Release> releases;
+
+  public PluginInfo(String id, List<Release> releases) {
+    this.id = id;
+    this.releases = releases;
+  }
+
+  static class Release {
```
jonsie, line 31: Should we add the release `state` field here?

PR #724, opened 2020-05-05
fix(events): Add dynamic config service to send events

marchello2000:
It's not currently possible to start/stop `igor` from sending events of completed builds/newly published images to `echo`.
It is useful to have this functionality because igor should always be "polling" to have a local cache of "triggered" builds/images.

This change allows an operator to:
* have two clusters of `igor` both polling but only one sending events (i.e. triggering pipeline)
* switch which cluster is sending events without generating duplicate events/double triggering

### DIFF
--- a/igor-core/src/main/java/com/netflix/spinnaker/igor/polling/CommonPollingMonitor.java
+++ b/igor-core/src/main/java/com/netflix/spinnaker/igor/polling/CommonPollingMonitor.java
@@ -22,6 +22,7 @@ import com.netflix.discovery.DiscoveryClient;
 import com.netflix.spectator.api.Id;
 import com.netflix.spectator.api.Registry;
 import com.netflix.spinnaker.igor.IgorConfigurationProperties;
+import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService;
 import com.netflix.spinnaker.kork.eureka.RemoteStatusChangedEvent;
 import java.time.Duration;
 import java.util.Optional;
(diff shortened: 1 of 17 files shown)


igor-core/src/main/java/com/netflix/spinnaker/igor/polling/CommonPollingMonitor.java
```diff
@@ -208,6 +219,8 @@ protected void internalPollSingle(PollContext ctx) {
             .set(0);
       }
 
+      sendEvents = sendEvents && isSendingEventsEnabled();
```
marchello2000, line 222: this is the actual change

igor-core/src/main/java/com/netflix/spinnaker/igor/polling/CommonPollingMonitor.java
```diff
@@ -264,6 +277,10 @@ public Long getLastPoll() {
     return lastPoll.get();
   }
 
+  private boolean isSendingEventsEnabled() {
+    return dynamicConfigService.isEnabled("igor.build.sendEvents", true);
```
ajordens, line 281: Not sure we need `igor` in here since the fp is going to be scoped to `igor` the application ... should the default be `true` or `false`?

```
igor:
  build:
    sendEvents: true
```

The `igor` in that config snippet would also seem weird in `igor.yml`.

We do have some config under `spinnaker.build` tho?

marchello2000: `spinnaker.build.sendEvents` seems a bit weird though... but it would fit with convention - will switch to that.
def want to default to `true`, since that's the behavior today

marchello2000: https://github.com/spinnaker/igor/pull/725

PR #725, opened 2020-05-06
fix(events): rename the send events prop

marchello2000:
Follow up to feedback in https://github.com/spinnaker/igor/pull/724.
Renaming the property to `spinnaker.build.sendEventsEnabled`, this way it is symmetric to existing properties in `igor`, e.g.:

```yaml
spinnaker:
  build:
    pollingEnabled: true
    sendEventsEnabled: false
```

### DIFF
--- a/igor-core/src/main/java/com/netflix/spinnaker/igor/polling/CommonPollingMonitor.java
+++ b/igor-core/src/main/java/com/netflix/spinnaker/igor/polling/CommonPollingMonitor.java
@@ -219,7 +219,7 @@ public abstract class CommonPollingMonitor<I extends DeltaItem, T extends Pollin
             .set(0);
       }
 
-      sendEvents = sendEvents && isSendingEventsEnabled();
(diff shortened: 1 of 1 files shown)


igor-core/src/main/java/com/netflix/spinnaker/igor/polling/CommonPollingMonitor.java
```diff
@@ -277,8 +277,8 @@ public Long getLastPoll() {
     return lastPoll.get();
   }
 
-  private boolean isSendingEventsEnabled() {
-    return dynamicConfigService.isEnabled("igor.build.sendEvents", true);
+  private boolean isSendEventsEnabled() {
+    return dynamicConfigService.getConfig(Boolean.class, "spinnaker.build.sendEventsEnabled", true);
```
marchello2000, line 281: i changed this to `getConfig` vs `isEnabled` to not have the extra `.enabled` tacked on, to make it consistent with the `isPollingEnabled` prop

PR #730, opened 2020-05-07
fix(decorator): Expect at least 3 fields in debian package name

adriengentil:
When a debian package name has no underscores in its name, the method
extractVersion throws the exception java.util.NoSuchElementException because it
cannot pop an empty list. When it happens Igor stops processing the repos
(Travis in my case) and the triggers.

This fix ensures the debian package name is composed of at least 3 fields in
order to enforce the naming convention name_version_architecture.deb


igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/artifact/decorator/DebDetailsDecorator.groovy
```diff
@@ -41,7 +41,8 @@ class DebDetailsDecorator implements ArtifactDetailsDecorator {
         if (!genericArtifact.fileName) {
             return false
         }
-        return genericArtifact.fileName.tokenize('.').last() == "deb"
+        return (genericArtifact.fileName.tokenize('.').last() == "deb" &&
+          genericArtifact.fileName.tokenize(versionDelimiter).size() >= 3)
```
cfieber, line 45: could you throw a comment in here that the deb file naming convention is package_version_architecture just so there is a bit of context why we are requiring 3 components

adriengentil: Done!

PR #740, opened 2020-05-12
refactor(gcb): Make GCB account immutable and null-safe

ezimanyi:
This commit moves the GoogleCloudBuildProperties.Account class to be immutable and use constructor binding. It also makes the minor change of storing the result of Permissions.build() instead of the raw builder, rather than calling Permissions.build() every time we ask for the permissions.

It also annotates the class as NonnullByDefault, and handles any necessary null defaulting in the constructor.

### DIFF
--- a/igor-web/src/main/java/com/netflix/spinnaker/igor/config/GoogleCloudBuildProperties.java
+++ b/igor-web/src/main/java/com/netflix/spinnaker/igor/config/GoogleCloudBuildProperties.java
@@ -16,13 +16,20 @@
 
 package com.netflix.spinnaker.igor.config;
 
+import com.google.common.base.Strings;
 import com.netflix.spinnaker.fiat.model.resources.Permissions;
 import com.netflix.spinnaker.igor.model.BuildServiceProvider;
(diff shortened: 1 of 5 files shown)


igor-web/src/main/java/com/netflix/spinnaker/igor/config/GoogleCloudBuildProperties.java
```diff
@@ -33,22 +40,37 @@
...
+    @Builder
+    @ConstructorBinding
+    @ParametersAreNullableByDefault
+    public Account(
+        String name,
+        String project,
+        String subscriptionName,
+        String jsonKey,
+        Permissions.Builder permissions) {
+      this.name = Strings.nullToEmpty(name);
```
cfieber, line 61: similar question to the other PR: are any of these fields required? should there be a constraint check rather than a nullToEmpty?

ezimanyi: Good point, addressed in #742.

PR #745, opened 2020-05-13
chore(docker): remove v1 docker tag migrator

marchello2000:
Pretty sure this isn't needed anymore since the migration took place in 2018


marchello2000:
@robzienert just want to make sure this is "safe" to delete :)

PR #750, opened 2020-05-14
chore(docker): remove v1 docker tag migrator

marchello2000:
Fix for a missed entry from https://github.com/spinnaker/igor/pull/745


marchello2000:
thanks groovy for compiling invalid code!

PR #762, opened 2020-05-27
fix(concourse): support old and new concourse auth

jaredstehler:
Fixes concourse authentication for clusters >= v6.1.0 by detecting version and providing different implementation accordingly.

Addresses issue: https://github.com/spinnaker/spinnaker/issues/5797


igor-web/src/main/java/com/netflix/spinnaker/igor/concourse/client/ConcourseClient.java
```diff
@@ -110,12 +146,20 @@ private void refreshTokenIfNecessary() {
...
+            : tokenServiceV2.passwordToken(
+                "password", user, password, "openid profile email federated:id groups");
+
     tokenExpiration = token.getExpiry();
     return token;
   }
 
+  public Response userInfo() {
+    return skyServiceV1 != null ? skyServiceV1.userInfo() : skyServiceV2.userInfo();
+  }
```
jervi, line 161: This feels a bit dirty to me. I'd consider creating a new `SkyService` interface that both V1 and V2 extends from, given that they both just contains the `userInfo()` method signature. Then you can have just a single `skyService` field, and not care about the implementation and possible `null`-values here.
And then do exactly the same with `TokenService`.

jervi: Example:
```
public interface SkyServiceV2 extends SkyService {
  @GET("/api/v1/user")
  @Override
  Response userInfo();
}
```

jaredstehler: fully agree! unfortunately, the library in use here, resteasy, doesn't allow extending interfaces :(

jaredstehler: see https://github.com/square/retrofit/issues/504 (appears this may have been fixed in a recent version; the one in use here is quite old)

jervi: Oh, I wasn't aware... Yuck. 😛
Maybe put a little comment in the code with that explanation?

donnyyung:
Would it be possible for us to cherry-pick this commit into 1.21 and 1.20 as well?

jaredstehler:
@Mergifyio backport release-1.20.x

PFacheris:
Noticed the above backport attempt failed and would also like to see this backported.

@robzienert + @jervi are you the appropriate people to ping for that?

kevinawoo:
FYI, we have a new [backport policy](https://spinnaker.io/community/contributing/releasing/#release-branch-patch-criteria) that only accepts regressions. There isn't a regression that the community introduced, but an external service changed.

Since we don't have any control of external services and we're supporting 3 minor versions, if concourse decided to change their auth and it broke a prior version, then Spinnaker as a service does have a regression and we'll need to backport the fix for it.

@mergifyio backport release-1.20.x release-1.21.x

kevinawoo:
@Mergifyio backport release-1.20.x

kevinawoo:
@Mergifyio backport release-1.21.x

PR #770, opened 2020-06-02
chore(travis): Move Travis integration to igor-monitor-travis

jervi:
Had to move some classes to `igor-core` to be able to do this, as both Jenkins and Travis (maybe other ci's as well?) depends on them. We could probably consider moving some of this to kork instead.
Also converted a bunch of classes to Java. No Groovy left in `igor-core` (except tests) \o/


jervi:
@robzienert Can you please take a look? 😃

jervi:
@spinnaker/oss-reviewers PTAL

igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/TravisCache.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2018 Netflix, Inc.
+ * Copyright 2020 Schibsted ASA.
```
robzienert, line 2: Mind the copyright changes that alter companies.

jervi: Oh, sorry. I did change a whole lot of them back, but seems this one slipped though the cracks. Bad IntelliJ!
I’ll fix it in the morning.

jervi:
> Are any of these things used by other services?
No, I don't think so. It used to be artifact related classes, but they're already extracted.

PR #773, opened 2020-06-02
fix(managed-delivery): Increase limit for delivery config size

luispollo:
The default limit of lines for text files retrieved using the Stash API is 500, and we came across files that were larger than that. This PR increases the limit to 2,000, which is not a definitive solution, but buys us some time until we can get back around to handling pagination.

Fixes https://github.com/spinnaker/keel/issues/1262.


igor-web/src/main/java/com/netflix/spinnaker/igor/scm/ManagedDeliveryScmService.java
```diff
@@ -63,10 +65,15 @@ public ManagedDeliveryScmService(
...
-            .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
+    this.jsonMapper = configureMapper(new ObjectMapper());
+    this.yamlMapper = configureMapper(new ObjectMapper(new YAMLFactory()));
+  }
+
+  private static ObjectMapper configureMapper(ObjectMapper mapper) {
+    return mapper
+        .disable(FAIL_ON_UNKNOWN_PROPERTIES)
+        .enable(ACCEPT_CASE_INSENSITIVE_ENUMS)
+        .setSerializationInclusion(NON_NULL);
```
luispollo, line 76: These are probably not needed. I copied them from the keel config, but here we just use deserialization into a `Map`, so they shouldn't matter. I decided to construct my own `ObjectMapper`s in this class because I was afraid configuration elsewhere might inadvertently break the parsing here.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/stash/client/StashMaster.groovy
```diff
@@ -26,8 +26,10 @@ import retrofit.RetrofitError
  */
 @Slf4j
 class StashMaster extends AbstractScmMaster {
-    StashClient stashClient
-    String baseUrl
+  public static final int DEFAULT_TEXT_FILE_MAX_LINES = 2000
```
luispollo, line 29: This should be a safe limit for now, is my assumption. I'm hoping to buy time so we can circle back to handling pagination later.

igor-web/src/main/java/com/netflix/spinnaker/igor/scm/ManagedDeliveryScmService.java
```diff
@@ -63,10 +61,8 @@ public ManagedDeliveryScmService(
     this.gitHubMaster = gitHubMaster;
     this.gitLabMaster = gitLabMaster;
     this.bitBucketMaster = bitBucketMaster;
-    this.jsonMapper = jsonMapper;
-    this.yamlMapper =
-        new ObjectMapper(new YAMLFactory())
-            .disable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES);
+    this.jsonMapper = new ObjectMapper();
+    this.yamlMapper = new ObjectMapper(new YAMLFactory());
```
luispollo, line 65: I decided to instantiate my own `ObjectMappers` here to avoid any unexpected influence from configuration elsewhere.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/stash/client/StashMaster.groovy
gal-yardeni: sorry to ask it here (I don't know the answer) - but how much work is it to add pagination? I'm afraid someone will have a huge config and then we will need to modify this again (and cut a release, deploy and so on).

luispollo: OK, I overcame my laziness and implemented it the right way. 🙂

gal-yardeni:
I don't see my previous comment here about pagination, but THANK YOU for handling it!

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/stash/client/StashMaster.groovy
```diff
@@ -44,9 +47,20 @@ class StashMaster extends AbstractScmMaster {
...
+      String contents = ""
+      boolean lastPage = false
+      int start = 0
+      while (!lastPage) {
+        log.debug("Retrieving text file contents from project: $projectKey, repo: $repositorySlug, path: $path, ref: $ref, start: $start")
+        TextLinesResponse response = stashClient.getTextFileContents(
+          projectKey, repositorySlug, path, ref, DEFAULT_PAGED_RESPONSE_LIMIT, start)
+        lastPage = response.isLastPage
+        start = response.start + response.size
+        contents += response.toTextContents() + "\n"
```
lorin, line 61: It's a bit more efficient to use a [StringBuilder](https://docs.oracle.com/javase/7/docs/api/java/lang/StringBuilder.html) instead of repeated string conatenation in a loop.

luispollo:
> I don't see my previous comment here about pagination, but THANK YOU for handling it!

They're still there! https://github.com/spinnaker/igor/pull/773#discussion_r434261814 🙂

PR #775, opened 2020-06-04
refactor(*): Replace Hystrix with Resilience4j

robzienert:
Only one place where Hystrix was actually used for fallbacks, which I migrated. Everything else was just removed.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/service/JenkinsService.java
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/service/JenkinsService.java
@@ -22,7 +22,6 @@ import static org.springframework.http.HttpStatus.NOT_FOUND;
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.netflix.spinnaker.fiat.model.resources.Permissions;
-import com.netflix.spinnaker.hystrix.SimpleJava8HystrixCommand;
 import com.netflix.spinnaker.igor.build.model.GenericBuild;
 import com.netflix.spinnaker.igor.build.model.GenericGitRevision;
 import com.netflix.spinnaker.igor.exceptions.ArtifactNotFoundException;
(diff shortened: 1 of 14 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/service/JenkinsService.java
```diff
@@ -91,10 +90,7 @@ private String encode(String uri) {
   }
 
   public ProjectsList getProjects() {
```
cfieber, line 92: The `grouppKey` in here is a distinct name per jenkins host, so that we would stop requests to a bad jenkins master. I don't know that we actually get much benefit from tripping that circuit breaker when there are problems, but I wonder what still uses groupKey in here (and other spots that got de-hystrix'd)

igor-web/src/main/java/com/netflix/spinnaker/igor/travis/service/TravisService.java
```diff
@@ -324,8 +310,9 @@ public Build getBuild(String repoSlug, int buildNumber) {
             error -> {
               log.warn("An error occurred while fetching new jobs from Travis.", error);
               return Collections.emptyList();
```
cfieber, line 312: this seems like a pretty questionable fallback but I don't have skin in the Travis game so I'm not proposing we change anything

jervi: It's the best fallback. Silent and sneaky!

But yeah, don't change it for now, but I'll try to remember to look into this.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/service/JenkinsService.java
robzienert: Just checked, `groupKey` isn't used anywhere else, I just failed to delete it.

I can add circuit breaking back into `JenkinsService` if we think it'd be useful. I suspect the only time it comes into play is when we take down a Jenkins master to upgrade - I think there's legitimacy there. WDYT?

robzienert: Added circuit breaking back in for jenkins.

PR #777, opened 2020-06-10
fix(resilience4j): Circuit breaker ignores 404s from Jenkins

robzienert:


ajordens:
LGTM

igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/service/JenkinsService.java
```diff
@@ -83,7 +84,18 @@ public JenkinsService(
...
     this.permissions = permissions;
-    this.circuitBreaker = circuitBreakerRegistry.circuitBreaker("jenkins-" + jenkinsHostId);
+    this.circuitBreaker =
+        circuitBreakerRegistry.circuitBreaker(
+            "jenkins-" + jenkinsHostId,
+            CircuitBreakerConfig.custom()
+                .ignoreException(
+                    (e) -> {
+                      if (e instanceof RetrofitError) {
+                        return ((RetrofitError) e).getResponse().getStatus() == 404;
```
ajordens, line 94: Maybe a `error.getKind() == HTTP && ...`

robzienert: Do we think we'd get a 404 status with any other Kind, or is this a bases covered sorta thing?

ajordens: Bases covered ... I think the `getResponse()` could be null for some other Kinds.

PR #811, opened 2020-07-09
fix(jenkins): Stop or delete from queue

marchello2000:
Stop a build can mean a) stopping it or b) removing it from the queue if hasn't started.
Today, igor, will attempt both if the build has been started. This can cause a 500 from jenkins server which is propagated to orca which then fails to actually complete the cancellation of a pipeline.
(it's mostly harmless but it can cause issues and it also generates a bunch of exception noise)

This change stops the build if it was started OR removes it from the queue if it wasn't started not both


gal-yardeni:
@marchello2000 is this related to the oncall question from yesterday? :)

marchello2000:
> @marchello2000 is this related to the oncall question from yesterday? :)

No, not related, just a general error handling improvement

PR #814, opened 2020-07-09
feat(discovery): refactor to use shared discovery abstraction from kork-core

cfieber:


cfieber:
this will break pending a kork release including https://github.com/spinnaker/kork/pull/714

cfieber:
closes #812

PR #821, opened 2020-07-21
feat(artifacts): Add type parameter to ArtifactService

luispollo:
Adds a `type` parameter to the `ArtifactService` interfaces methods such that we can query different types of artifacts (not just Debian).

### DIFF
--- a/igor-web/src/main/java/com/netflix/spinnaker/igor/artifacts/ArtifactController.java
+++ b/igor-web/src/main/java/com/netflix/spinnaker/igor/artifacts/ArtifactController.java
@@ -46,18 +46,20 @@ public class ArtifactController {
   public List<String> getVersions(
       @PathVariable("provider") String provider,
       @PathVariable("name") String name,
-      @RequestParam(value = "releaseStatus", required = false) String releaseStatus) {
+      @RequestParam(value = "releaseStatus", required = false) String releaseStatus,
+      @RequestParam(value = "type", required = false, defaultValue = "deb") String type) {
     ArtifactService artifactService = getService(provider);
-    return artifactService.getArtifactVersions(name, releaseStatus);
+    return artifactService.getArtifactVersions(type, name, releaseStatus);
   }
 
   @GetMapping("/{provider}/{name}/{version:.+}")
   public Artifact getArtifact(
       @PathVariable("provider") String provider,
       @PathVariable("name") String name,
-      @PathVariable("version") String version) {
+      @PathVariable("version") String version,
+      @RequestParam(value = "type", required = false, defaultValue = "deb") String type) {
     ArtifactService artifactService = getService(provider);
-    return artifactService.getArtifact(name, version);
+    return artifactService.getArtifact(type, name, version);
   }
 
   private ArtifactService getService(String serviceName) {
--- a/igor-web/src/main/java/com/netflix/spinnaker/igor/artifacts/ArtifactController.java
+++ b/igor-web/src/main/java/com/netflix/spinnaker/igor/artifacts/ArtifactController.java
@@ -46,10 +46,10 @@ public class ArtifactController {
   public List<String> getVersions(
       @PathVariable("provider") String provider,
       @PathVariable("name") String name,
-      @RequestParam(value = "releaseStatus", required = false) String releaseStatus,
+      @RequestParam(value = "releaseStatus", required = false) List<String> releaseStatuses,
       @RequestParam(value = "type", required = false, defaultValue = "deb") String type) {
     ArtifactService artifactService = getService(provider);
-    return artifactService.getArtifactVersions(type, name, releaseStatus);
+    return artifactService.getArtifactVersions(type, name, releaseStatuses);
   }
 
   @GetMapping("/{provider}/{name}/{version:.+}")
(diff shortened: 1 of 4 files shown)


igor-web/src/main/java/com/netflix/spinnaker/igor/artifacts/ArtifactController.java
```diff
@@ -46,18 +46,20 @@ public ArtifactController(ArtifactServices artifactServices) {
   public List<String> getVersions(
       @PathVariable("provider") String provider,
       @PathVariable("name") String name,
-      @RequestParam(value = "releaseStatus", required = false) String releaseStatus) {
+      @RequestParam(value = "releaseStatus", required = false) String releaseStatus,
+      @RequestParam(value = "type", required = false, defaultValue = "deb") String type) {
```
luispollo, line 50: I added this as a request param and not a path variable because it would cause a conflict with between the path patterns in `getArtifact` and `getVersions`.

igor-web/src/main/java/com/netflix/spinnaker/igor/artifacts/ArtifactService.java
```diff
@@ -33,9 +33,26 @@
...
   /** Used to populate the manual trigger dropdown with options */
   @Nonnull
-  List<String> getArtifactVersions(@Nonnull String name, String releaseStatus);
+  List<String> getArtifactVersions(
+      @Nonnull String type, @Nonnull String name, String releaseStatus);
 
   /** Used to fetch a specific artifact for decorating a trigger */
   @Nonnull
-  Artifact getArtifact(@Nonnull String name, @Nonnull String version);
+  Artifact getArtifact(@Nonnull String type, @Nonnull String name, @Nonnull String version);
```
emjburns, line 41: In the future I'd prefer if you added a new argument as the last parameter to this function. No need to change here though.

igor-web/src/main/java/com/netflix/spinnaker/igor/artifacts/ArtifactService.java
```diff
@@ -33,9 +33,26 @@
...
+
+  /**
+   * Used to populate the manual trigger dropdown with options (backwards-compatibility for Debian)
```
emjburns, line 44: Why do you need these backwards compatibility functions when you default the parameter in the controller?

igor-web/src/main/java/com/netflix/spinnaker/igor/artifacts/ArtifactService.java
luispollo: Can you elaborate on why? I added it to that position specifically because it made more logical sense to me that you'd specify the type before the name.

igor-web/src/main/java/com/netflix/spinnaker/igor/artifacts/ArtifactService.java
luispollo: Good point. I started from the bottom up, so I added these here in case there were any other users beyond that controller, but I think there are none, so I can probably remove them safely.

igor-web/src/main/java/com/netflix/spinnaker/igor/artifacts/ArtifactService.java
emjburns: Groovy doesn't always tell you when you break function calls, because groovy sucks. We have a lot of groovy, so I'm just always cautious (not applicable here, but applicable elsewhere). Also, I think it's easy to mix up test parameters when you add at the beginning.

luispollo: ACK.

igor-web/src/main/java/com/netflix/spinnaker/igor/artifacts/ArtifactController.java
```diff
@@ -46,18 +46,20 @@ public ArtifactController(ArtifactServices artifactServices) {
   public List<String> getVersions(
       @PathVariable("provider") String provider,
       @PathVariable("name") String name,
-      @RequestParam(value = "releaseStatus", required = false) String releaseStatus) {
+      @RequestParam(value = "releaseStatus", required = false) List<String> releaseStatuses,
```
luispollo, line 49: This was previously limiting the ability to query artifacts matching multiple statuses.

igor-web/src/main/java/com/netflix/spinnaker/igor/artifacts/ArtifactService.java
```diff
@@ -33,9 +33,10 @@
 
   /** Used to populate the manual trigger dropdown with options */
   @Nonnull
-  List<String> getArtifactVersions(@Nonnull String name, String releaseStatus);
+  List<String> getArtifactVersions(
+      @Nonnull String type, @Nonnull String name, List<String> releaseStatuses);
```
luispollo, line 37: This was previously limiting the ability to query artifacts with multiple statuses.

igor-web/src/main/java/com/netflix/spinnaker/igor/artifacts/ArtifactController.java
emjburns: Is this change backwards compatible with the manual trigger dropdown only sending one status?

luispollo: Yes, because the way this is encoded in query parameters when it's a list is to have multiple occurrences of `releaseStatus=<status>` in the query string. This works for 1 to N.

emjburns: love it, so great!

PR #826, opened 2020-07-27
fix(concourse): handle missing branch in git metadata

jaredstehler:
Sets branch name to the sha1 short-ref if missing from concourse metadata

fixes https://github.com/spinnaker/spinnaker/issues/5942


jaredstehler:
done

PR #828, opened 2020-08-01
feat(managed-delivery): Retrieve Managed Delivery manifests from GitHub

kzwang:
Add support for retrieving managed delivery manifests from GitHub repos.

Also fixed GitHub deprecated API authentication through query parameters (https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param/)

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/github/client/GitHubMaster.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/github/client/GitHubMaster.groovy
@@ -16,53 +16,63 @@
 
 package com.netflix.spinnaker.igor.scm.github.client
 
-import com.netflix.spinnaker.igor.config.GitHubProperties
-import com.netflix.spinnaker.igor.scm.AbstractScmMaster
-import com.netflix.spinnaker.retrofit.Slf4jRetrofitLogger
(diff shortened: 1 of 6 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/github/client/GitHubMaster.groovy
```diff
@@ -16,53 +16,63 @@
...
+import groovy.util.logging.Slf4j
+import retrofit.RetrofitError
 
-import javax.validation.Valid
+import java.util.stream.Collectors
 
 /**
  * Wrapper class for a collection of GitHub clients
  */
-class GitHubMaster  extends AbstractScmMaster {
```
gal-yardeni, line 34: since it's in OSS and we don't really know who is using this class, can it be backward-incompatible?

kzwang: The current `GitHubMaster` is just a copy of `GitHubConfig`, that defines the `GitHubMaster` bean, which is not used, and it didn't implement any methods from `AbstractScmMaster`, so shouldn't have backward compatibility issues.

gal-yardeni:
@emjburns @luispollo appreciate if you could take a look whenever you can!

PR #838, opened 2020-08-19
chore(*): Use SprintBootApplication annotation on Main class

jonsie:
This annotation is useful as a way of consistently obtaining the Main class package in our services so we can determine package information, like version, at runtime.


luispollo:
@jonsie Side-question: how about the `Implementation-Version` metadata for the package? Where does that come from?

jonsie:
@luispollo That is set at build time.  We pass it in as the parameter `ossVersion` and it is set via Gradle: https://github.com/spinnaker/spinnaker-gradle-project/blob/8063c55cb73c9cf888ef1c30973d4edfd0b232aa/spinnaker-project-plugin/src/main/groovy/com/netflix/spinnaker/gradle/baseproject/SpinnakerBaseProjectConventionsPlugin.groovy#L48-L54

PR #839, opened 2020-08-20
fix(codebuild): set region when creating sts client

Kaixiang-AWS:
There's a race condition where default region could not be found, causing
STS client failed to be created. To fix it, we set STS client region to
what specified in the codebuild halyard config.

Fixed https://github.com/spinnaker/spinnaker/issues/5974


german-muzquiz:
@Kaixiang-AWS @clareliguori in order for a PR to be backported to previous releases, it must meet the [documented criteria](https://spinnaker.io/community/contributing/releasing/#release-branch-patch-criteria). I see that the following is missing:
* It must fix something that was already working, and broke in a previous release (a regression).
* It needs to add test coverage validating the regression and the fix.

I see that this issue may have been there since the beginning, meaning that it doesn't fix a regression. Can you confirm?

german-muzquiz:
I understand that this is a bug, but at this point only fixes for regressions and security vulenrabilities are allowed into release branches. Until that policy is changed, non-regression bug fixes will need to wait for the next release.

atyutyunnik:
1.21 is the most current release. I honestly expected the Spinnaker dev team to be more...agile, not putting a policy/process over customer's needs, a policy that makes little sense in this context. 1.23 is not even scheduled

clareliguori:
@german-muzquiz I missed seeing the updated release branch criteria.  You are correct, this is not a regression.

PR #844, opened 2020-08-24
feat(ciService): add more fields to generic ci service

gal-yardeni:
- adding build number and commit id to generic CI service endpoint
- adding `NoArgsConstructor` and `AllArgsConstructor` to `GenericBuild`

### DIFF
--- a/igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericBuild.java
+++ b/igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericBuild.java
@@ -26,6 +26,8 @@ import lombok.Data;
 import lombok.NoArgsConstructor;
 
 @Data
+@NoArgsConstructor
+@AllArgsConstructor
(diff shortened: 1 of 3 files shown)


igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericBuild.java
```diff
@@ -26,6 +26,8 @@
 import lombok.NoArgsConstructor;
 
 @Data
+@NoArgsConstructor
```
gal-yardeni, line 29: hi @robzienert do you have any concerns with adding these annotations here?

robzienert: No, but I'm not sure it's necessary? `@Data` already implicitly adds `@RequiredArgsConstructor`.

robzienert: Oh. Nvm.

PR #596, opened 2020-01-22
feat(jenkins): Support for adding custom KeyStore for Jenkins clients

jervi:
```
jenkins:
  masters:
    - address: "https://jenkins.com"
      name: spinnaker
      keyStore: /path/to/keystore.p12
      keyStoreType: PKCS12
      keyStorePassword: my_keystore_password
```
or
```
jenkins:
  masters:
    - address: "https://jenkins.com"
      name: spinnaker
      keyStore: /path/to/keystore.jks
      keyStorePassword: my_keystore_password
```

PR #856, opened 2020-09-09
fix(jenkins): Igor fails to start if only keystore is configured

jervi:
This configuration worked fine in okhttp2, but not in okhttp3.
As a workaround, Igor will try to use the keystore as a truststore as well, and prints a warning in the log.

"Bug" introduced in https://github.com/spinnaker/igor/pull/834. Keystore support was added in https://github.com/spinnaker/igor/pull/596.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/JenkinsConfig.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/JenkinsConfig.groovy
@@ -147,7 +147,7 @@ class JenkinsConfig {
             if (host.trustStore.equals("*")) {
                 trustManagers = [new TrustAllTrustManager()]
             } else {
-                trustManagers = createTrustStore(host.trustStorePassword, host.trustStoreType, host.trustStore)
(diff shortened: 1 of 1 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/JenkinsConfig.groovy
```diff
@@ -170,13 +161,19 @@ class JenkinsConfig {
             keyManagerFactory.init(keyStore, keyStorePassword.toCharArray())
 
             keyManagers = keyManagerFactory.keyManagers
+
+            if (trustManagers == null) {
+                log.warn("${host.name}: okhttp3 (unlike okhttp2) doesn't support configuring only a keystore without " +
+                        "a truststore. Please configure a truststore to get rid of this message. Trying to use the " +
+                        "keystore '${host.keyStore}' as a truststore as well. Your mileage may vary.")
```
robzienert, line 168: Ah, I'm not sure we want to be advising this.

robzienert (changes requested):
Sketchy advise in logs, I'd get rid of that.

robzienert (approved):
Oh. I read the log backwards. Looks fine.

PR #863, opened 2020-09-18
fix(jenkins): Don't kick off multiples of the same build

marchello2000:
The issue here is that when an operation takes a long time (e.g. talking to jenkins to kick off a build)
the prolonged time can cause `orca` to try to resubmit the request multiple times thus kicking off multiple (or infinite number of) builds.

We solve this, by setting a flag that a given request (keyed of X-SPINNAKER-REQUEST-ID and master/job/param) is being processed so if an identical request comes in the controller can detect it and respond with `202 ACCEPTED`. This will cause orca to retry (see https://github.com/spinnaker/orca/pull/3915).
When the job is finally kicked off, the resulting build number will be stored in the cache and next time orca queries it will get the build number back


marchello2000:
IntelliJ decided to reformat the spacing.. so review with white space off : https://github.com/spinnaker/igor/pull/863/files?diff=split&w=1
Also, here is corresponding orca PR (https://github.com/spinnaker/orca/pull/3915)

igor-web/src/main/groovy/com/netflix/spinnaker/igor/build/BuildController.groovy
```diff
@@ -51,205 +55,238 @@ import static org.springframework.http.HttpStatus.NOT_FOUND
...
+    @RequestParam Map<String, String> requestParams, HttpServletRequest request) {
+    def job = ((String) request.getAttribute(
+      HandlerMapping.PATH_WITHIN_HANDLER_MAPPING_ATTRIBUTE)).split('/').drop(4).join('/')
+
+    String pendingKey = computePendingBuildKey(master, job, requestParams)
+    String buildNumber = null
+
+    PendingOperationsCache.OperationState pendingStatus = pendingOperationsCache.getAndSetOperationStatus(pendingKey, PendingOperationsCache.OperationStatus.PENDING, "")
+    if (pendingStatus.status == PendingOperationsCache.OperationStatus.PENDING) {
+      throw new BuildOperationPendingException(master, job)
```
jonsie, line 186: Throwing this exception like this initially struck me as odd - curious for the reasoning?  Instead of throwing an exception is it possible to just return an accepted response with a message?

marchello2000: Yeah, I actually hate it - but I couldn't figure out how to return `202-ACCEPTED` if you can point me in the right direction i would love to change it as it will reduce the noise in logs

marchello2000: nvm, @srekapalli put me on the path of righteousness :)

marchello2000: fixed

srekapalli (approved):
LGTM

marchello2000:
> LGTM

thank you!

PR #866, opened 2020-09-22
feat(date): adding date to docker images metadata, and log statement with timing

gal-yardeni:
fetching `date` field from docker metadata, and adding a log statement which will help us figure out the time difference between including all details from the registry to just include some of them.

### DIFF
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerMonitor.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerMonitor.groovy
@@ -101,10 +101,10 @@ class DockerMonitor extends CommonPollingMonitor<ImageDelta, DockerPollingDelta>
         Set<String> cachedImages = cache.getImages(account)
 
         long startTime = System.currentTimeMillis()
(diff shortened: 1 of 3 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/docker/DockerMonitor.groovy
```diff
@@ -101,7 +101,11 @@ class DockerMonitor extends CommonPollingMonitor<ImageDelta, DockerPollingDelta>
         Set<String> cachedImages = cache.getImages(account)
 
         long startTime = System.currentTimeMillis()
-        List<TaggedImage> images = AuthenticatedRequest.allowAnonymous { dockerRegistryAccounts.service.getImagesByAccount(account) }
+        List<TaggedImage> images = AuthenticatedRequest.allowAnonymous { dockerRegistryAccounts.service.getImagesByAccount(account, true) }
```
emjburns, line 104: Let's do this in two separate changes - first, roll out the timing. Then, change the API call. That way we get a good indication of if the call takes longer

gal-yardeni: done!

PR #870, opened 2020-09-24
chore(dependencies): Autobump korkVersion

spinnakerbot:


jonsie:
Looking at it.

jonsie:
Bumping here: https://github.com/spinnaker/igor/pull/871

PR #878, opened 2020-10-02
Do not force nodeId configured when checking nexus webhook

jmalvarezf-lmes:
This PR addresses a problem that we do have with Nexus repositories. When you configure a nexus repository, right now you are forced to use nodeId, although it is optional, because when you configure webhooks in Nexus3, nodeId is always present. To allow this property to be really optional, we propose this PR, that checks previously if there are Nexus repositories configured with nodeId, and if there is none, check the repo name instead.

Hope you find it useful.


gal-yardeni:
hi @jmalvarezf-lmes! overall it looks good to me, any chance you can add a test for it? thanks!

jmalvarezf-lmes:
Hi @gal-yardeni!

Sure, I'll do it tomorrow.

Thanks,

jmalvarezf-lmes:
Test added!

gal-yardeni (approved):
lgtm

PR #894, opened 2020-10-15
fix(retrofit): Use the retrofit log level bean from retrofit configuration in kork-web

marchello2000:
Make it more consistent logging wise (`echo` client had logging set to NONE 😱 )


srekapalli (approved):

ajordens:
LGTM

PR #904, opened 2020-12-01
feat(buildLog): adding build log hook

gal-yardeni:
As a part of our effort to open source the CI view in Spinnaker, we have 2 endpoints which power this view:
1. get all builds, based on a project and repo names.
2. for each build, get the build log.

We already added the first hook (#1). This PR is adding the second hook to fetch the build log.
In order to use these hooks, you'll have to implement a `CiBuildService` in your `igor` fork, with concrete implementation to those two endpoints.

The CI view will call those endpoints from `deck` (the UI) via `gate`.


gal-yardeni:
cc: @emccallum

emjburns (approved):

PR #920, opened 2021-01-21
chore(config): Fix outdated example in igor-config

jervi:


gal-yardeni:
@jervi can you elaborate a bit on why you changed it to 100? thanks!

jervi:
I just updated it to reflect the default in Igor. The old `numberOfRepositories` setting is ignored by Igor - and did something else. It controlled how many recently built repos Igor would poll for changes (and it would then go on and poll each of these for recently built jobs). When I rewrote the Travis integration I opted to instead use another, simpler endpoint that just polls for recently built jobs across all repos - drastically reducing the number of requests to Travis. `numberOfJobs` controls how many jobs Travis should return per polling window.
https://github.com/spinnaker/igor/blob/master/igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/config/TravisProperties.java#L62-L64

gal-yardeni (approved):

PR #929, opened 2021-03-08
feat(CiBuild): extend Jenkins / Docker / Github clients with REST calls

nimakaviani:
- add extra necessary fields to Git and Build data structure for CiBuild support
- add calls to jenkins and github to pull commit details and build info
- add property field to enable CiBuild integration

/cc @luispollo 
/cc @gal-yardeni

### DIFF
--- a/igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericArtifact.java
+++ b/igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericArtifact.java
@@ -41,6 +41,14 @@ public class GenericArtifact {
     this.name = name;
     this.version = version;
     this.reference = reference;
+    this.fileName = String.format("%s:%s", name, version);
+  }
+
+  public GenericArtifact withUrl(String registry) {
+    String url = String.format("https://%s/v2/%s/blobs/%s", registry, name, reference);
+    this.displayPath = url;
+    this.url = url;
+    return this;
   }
 
   private String fileName;
@@ -50,6 +58,7 @@ public class GenericArtifact {
   private String name;
   private String type;
   private String version;
+  private String url;
   private Map<String, String> metadata;
   private boolean decorated;
 }
--- a/igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericArtifact.java
+++ b/igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericArtifact.java
@@ -44,8 +44,7 @@ public class GenericArtifact {
     this.fileName = String.format("%s:%s", name, version);
   }
 
-  public GenericArtifact withUrl(String registry) {
-    String url = String.format("https://%s/v2/%s/blobs/%s", registry, name, reference);
+  public GenericArtifact withUrl(String url) {
     this.displayPath = url;
     this.url = url;
     return this;
--- a/igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericArtifact.java
+++ b/igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericArtifact.java
@@ -22,6 +22,7 @@ import lombok.Builder;
 import lombok.Data;
 import lombok.NoArgsConstructor;
 import lombok.ToString;
+import lombok.experimental.Wither;
 
 @ToString
 @Data
@@ -29,6 +30,7 @@ import lombok.ToString;
 @AllArgsConstructor
 @Builder(toBuilder = true)
 @JsonInclude(JsonInclude.Include.NON_NULL)
+@Wither
 public class GenericArtifact {
   public GenericArtifact(String fileName, String displayPath, String relativePath) {
     this.fileName = fileName;
@@ -42,12 +44,6 @@ public class GenericArtifact {
     this.version = version;
     this.reference = reference;
     this.fileName = String.format("%s:%s", name, version);
-  }
-
-  public GenericArtifact withUrl(String url) {
-    this.displayPath = url;
-    this.url = url;
-    return this;
   }
 
   private String fileName;
--- a/igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericBuild.java
+++ b/igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericBuild.java
@@ -49,6 +49,21 @@ public class GenericBuild {
 
   private Map<String, ?> properties;
 
+  public GenericBuild withGitInfo(List<GenericGitRevision> revs) {
+    genericGitRevisions = revs;
+    return this;
+  }
+
+  public GenericBuild withProperties(Map<String, ?> properties) {
+    this.properties = properties;
+    return this;
+  }
+
+  public GenericBuild withArtifacts(List<GenericArtifact> artifacts) {
+    this.artifacts = artifacts;
+    return this;
+  }
+
   @Data
   @AllArgsConstructor
   @NoArgsConstructor
--- a/igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericBuild.java
+++ b/igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericBuild.java
@@ -24,11 +24,13 @@ import java.util.Map;
 import lombok.AllArgsConstructor;
 import lombok.Data;
 import lombok.NoArgsConstructor;
+import lombok.experimental.Wither;
 
 @Data
 @NoArgsConstructor
 @AllArgsConstructor
 @JsonInclude(JsonInclude.Include.NON_NULL)
+@Wither
 public class GenericBuild {
   private boolean building;
   private String fullDisplayName;
@@ -48,21 +50,6 @@ public class GenericBuild {
   private List<GenericGitRevision> genericGitRevisions;
 
   private Map<String, ?> properties;
-
-  public GenericBuild withGitInfo(List<GenericGitRevision> revs) {
-    genericGitRevisions = revs;
-    return this;
-  }
-
-  public GenericBuild withProperties(Map<String, ?> properties) {
-    this.properties = properties;
-    return this;
-  }
-
-  public GenericBuild withArtifacts(List<GenericArtifact> artifacts) {
-    this.artifacts = artifacts;
-    return this;
-  }
 
   @Data
   @AllArgsConstructor
--- a/igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericGitRevision.java
+++ b/igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericGitRevision.java
@@ -36,20 +36,4 @@ public class GenericGitRevision {
   private String message;
   private Instant timestamp;
   private String remoteUrl;
-
-  public GenericGitRevision withAuthorInfo(String committer, String message) {
-    this.committer = committer;
-    this.message = message;
-    return this;
-  }
-
-  public GenericGitRevision withTimestamp(Instant timestamp) {
-    this.timestamp = timestamp;
-    return this;
-  }
-
-  public GenericGitRevision withCompareUrl(String link) {
-    this.compareUrl = link;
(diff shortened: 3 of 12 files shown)


nimakaviani:
Just to clarify the purpose of this PR, the majority of the work here involves building hooks and adding REST calls that we use in the soon-to-be-released CiBuild plugin. The majority of the testing for the added functionality, we will add to the plugin.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/JenkinsConfig.groovy
```diff
@@ -128,13 +128,15 @@ class JenkinsConfig {
...
-                                       RestAdapter.LogLevel retrofitLogLevel,
-                                       int timeout = 30000) {
+  static JenkinsClient jenkinsClient(JenkinsProperties.JenkinsHost host,
+                                     JenkinsOkHttpClientProvider jenkinsOkHttpClientProvider = null,
+                                     JenkinsRetrofitRequestInterceptorProvider jenkinsRetrofitRequestInterceptorProvider = null,
+                                     Registry registry = null,
+                                     RestAdapter.LogLevel retrofitLogLevel = RestAdapter.LogLevel.BASIC,
+                                     int timeout = 30000){
 
+        OkHttpClient client = (jenkinsOkHttpClientProvider != null) ? jenkinsOkHttpClientProvider.provide(host) : new OkHttpClient()
```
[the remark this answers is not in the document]
luispollo: Never mind. I see this was already there before, just below.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/github/client/GitHubMaster.groovy
```diff
@@ -75,4 +76,12 @@ class GitHubMaster extends AbstractScmMaster {
     }
   }
 
+  @Override
+  Commit getCommitDetails(String projectKey, String repositorySlug, String sha) {
+    try {
+      return gitHubClient.commitInfo(projectKey, repositorySlug, sha)
+    }catch (Exception e) {
+      throw e
+    }
```
luispollo, line 85: Since this is just rethrowing any exceptions, it could be simplified to:
```suggestion
    return gitHubClient.commitInfo(projectKey, repositorySlug, sha)
```

igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericArtifact.java
```diff
@@ -41,6 +41,14 @@ public GenericArtifact(String type, String name, String version, String referenc
     this.name = name;
     this.version = version;
     this.reference = reference;
+    this.fileName = String.format("%s:%s", name, version);
+  }
+
+  public GenericArtifact withUrl(String registry) {
+    String url = String.format("https://%s/v2/%s/blobs/%s", registry, name, reference);
```
luispollo, line 48: This class is a `GenericArtifact`, but this URL looks like it's GitHub-specific. Can you please explain how this will be used?

luispollo:
> Just to clarify the purpose of this PR, the majority of the work here involves building hooks and adding REST calls that we use in the soon-to-be-released CiBuild plugin. The majority of the testing for the added functionality, we will add to the plugin.

Hi @nimakaviani. Just curious -- did you end up implementing an actual Spinnaker plugin for Igor as we had discussed? If there's code for it in GitHub already (even if not released yet), it might be helpful to link it in this PR for context.

igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericArtifact.java
nimakaviani: good catch. fixed it.

nimakaviani:
Thanks @emjburns for taking the time. The PR only introduces some minor refactoring and adds a few setters and some extra retrofit rest calls to JenkinsClient and DockerClient, following the existing patterns. By data, do you mean the payload from the REST API endpoints for Jenkins and DockerHub?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/scm/github/client/GitHubMaster.groovy
nimakaviani: fixed!

igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericGitRevision.java
```diff
@@ -36,4 +36,20 @@
...
+    this.message = message;
+    return this;
+  }
+
+  public GenericGitRevision withTimestamp(Instant timestamp) {
+    this.timestamp = timestamp;
+    return this;
+  }
+
+  public GenericGitRevision withLink(String link) {
```
luispollo, line 51: Consider renaming for consistency with what the method is changing:
```suggestion
  public GenericGitRevision withCompareUrl(String link) {
```

igor-web/src/main/java/com/netflix/spinnaker/igor/scm/AbstractScmMaster.java
```diff
@@ -32,4 +32,8 @@ public String getTextFileContents(
   public String getTextFileContents(String projectKey, String repositorySlug, String path) {
     return getTextFileContents(projectKey, repositorySlug, path, ScmMaster.DEFAULT_GIT_REF);
   }
+
+  public Object getCommitDetails(String projectKey, String repositorySlug, String commitSha) {
```
luispollo, line 36: Could this return a `Commit` object instead?

igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericGitRevision.java
nimakaviani: fixed

igor-web/src/main/java/com/netflix/spinnaker/igor/scm/AbstractScmMaster.java
nimakaviani: problem is, `Commit` is implemented per SCM, so there is no generic struct to return here unless we create an abstract one. Alternatively we can try to consolidate them all into something more generic, but that appeared to be a chore we can deal with independent of this PR.

igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericGitRevision.java
```diff
@@ -36,4 +36,20 @@
...
+
+  public GenericGitRevision withTimestamp(Instant timestamp) {
+    this.timestamp = timestamp;
+    return this;
+  }
+
+  public GenericGitRevision withCompareUrl(String link) {
+    this.compareUrl = link;
+    return this;
+  }
```
jervi, line 54: These two methods can be removed, they are automatically generated by the Lombok `@Wither` annotation.

igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericBuild.java
```diff
@@ -49,6 +49,21 @@
...
+
+  public GenericBuild withProperties(Map<String, ?> properties) {
+    this.properties = properties;
+    return this;
+  }
+
+  public GenericBuild withArtifacts(List<GenericArtifact> artifacts) {
+    this.artifacts = artifacts;
+    return this;
+  }
```
jervi, line 65: IMHO these methods should be removed and instead added using the Lombok annotation `@Wither`. You can probably just add it for the whole class, but if you only want these three methods, you can add it directly on the corresponding fields.

igor-web/src/main/java/com/netflix/spinnaker/igor/scm/AbstractScmMaster.java
jervi: What about using a `GenericGitRevision`? That more or less corresponds to a commit.

igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericArtifact.java
```diff
@@ -41,6 +41,13 @@ public GenericArtifact(String type, String name, String version, String referenc
     this.name = name;
     this.version = version;
     this.reference = reference;
+    this.fileName = String.format("%s:%s", name, version);
+  }
+
+  public GenericArtifact withUrl(String url) {
+    this.displayPath = url;
+    this.url = url;
+    return this;
```
jervi, line 50: Replace with Lombok's `@Wither` annotation.

jervi (commented):
Looks good, I just have some minor comments regarding Lombok

igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericArtifact.java
luispollo: In general, I agree with the suggestion, just note that @nimakaviani is actually setting two different fields within this method.

luispollo: In general, I agree with the suggestion, just note that @nimakaviani is actually setting two different fields within this method (which maybe an anti-pattern, I guess).

igor-web/src/main/java/com/netflix/spinnaker/igor/scm/AbstractScmMaster.java
luispollo: @nimakaviani I see. I'm OK tackling this in a separate PR.

igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericGitRevision.java
nimakaviani: good call. didnt notice `@Wither`. removed.

igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericBuild.java
nimakaviani: added `@Wither` to the whole class, doesn't seem to hurt. thanks!

igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericArtifact.java
nimakaviani: yeah it is setting two fields. will keep it as is.

jervi: I think that is a bad idea, as you're breaking the wither-pattern. If you really want to set two fields in the same with-method, then call it something else.

igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericGitRevision.java
nimakaviani: one little annoyance with `@Wither` however is that it doesnt do in-place update of the properties, so I had to capture the returned value.

igor-core/src/main/java/com/netflix/spinnaker/igor/build/model/GenericArtifact.java
nimakaviani: yeah fair point. on a second look, it made set them separately.

igor-web/src/main/java/com/netflix/spinnaker/igor/scm/AbstractScmMaster.java
nimakaviani: @jervi `GenericGitRevision` is close, but `commit` and `author` info returned from different SCMs (github or gitlab) are more complex data structures that do not match one-to-one to the stringified values in `GenericGitRevision`, unfortunately.

luispollo (approved):
LGTM.

PR #938, opened 2021-03-22
chore(release): Add weekly release

luispollo:
Similarly to what we've done for Keel recently, this PR adds a new GitHub action to automatically trigger a weekly release for Igor such that we don't have to manually cut releases. This is intended to enable other organizations consuming releases outside of Netflix.


nimakaviani (approved):
LGTM thanks!

PR #939, opened 2021-04-01
fix(jenkins): Using null breaks the cache mechanism

JHeilCoveo:
Fix issue https://github.com/spinnaker/spinnaker/issues/6217.

Here is what happens now :

* buildNumber gets casted from null to "null" when calling `pendingOperationsCache.setOperationStatus(pendingKey, PendingOperationsCache.OperationStatus.COMPLETED, buildNumber)`
* Next call from orca evaluate `!Strings.isNullOrEmpty(pendingStatus.value)` as true and returns "null" as job identifier thus breaking the next step.


dreynaud (approved):
Thanks for the fix!

PR #964, opened 2021-09-20
fix(travis): Avoid a potential NPE when retrieving latest builds

jervi:
After upgrading from Travis Enterprise 2 to 3, we ran into an issue where the build was `null` for some jobs. I believe this  is a bug in Travis, but I'm adding a check here to just avoid a NullPointerException that blocks the Travis provider from fetching builds if this issue should ever occur again.


jervi:
@spinnaker/oss-reviewers PTAL :)

dbyron-sf (approved):

PR #986, opened 2022-01-26
Implemented Gitlab CI polling and build property support

linjmeyer:
GitLab CI support has been requested and partially worked on a few times (e.g. [2047](https://github.com/spinnaker/spinnaker/issues/2047)).  There was some existing code for GitLab CI polling support, this PR adds additional features, major performance improvements and adds build property support.

## Changes to Existing Code

* Performance fixes in Gitlab polling
* Filter out Gitlab projects without CI enabled when polling for new builds
* Refactor to use list APIs whenever possible to avoid single API calls for many projects, pipelines and jobs
* Refactor to use Project Id, Pipeline Id and Job Ids instead of friendly names (e.g. team/project format).  This reduces API calls needed in some cases.

Existing code did support Gitlab CI polling and sending events to Echo however on our Gitlab CI instance this polling never completed after ~10 minutes so I stopped it.  The Gitlab APIs aren't great for the way Spinnaker polls CI providers.  This PR greatly reduces the number of calls made to GitLab CI which for our internal CI brought the poll time to around 10 seconds.  That did require removing some fields such as `pipeline.finishedAt` which would normally be a timestamp of when the Pipeline completed, however this field is not available using list APIs (`createdAt` property is available in list API instead).  I felt the performance improvements were worth giving up some minor build details.

## Additional features

* Build Properties support.  This is implemented using Gitlab Job logs by re-using Travis CI's log parser (more on Gitlab artifacts later)
* Basic documentation on adding GitLab CI masters

## Gitlab Artifacts

Gitlab does support artifacts and has an API endpoint for downloading them, however the API requires the full path to the file.  When Gitlab CI archives artifacts it keeps the folder structure, so you end up with something like `job/artifacts/whatever/hello.yaml`.  There is no way to list the individual files using the Gitlab APIs, so you need to know the full path in order to download it.  That unfortunately is incompatible with the current `BuildController.getProperties` method which has a path of `/builds/properties/{buildNumber}/{fileName}/{master:.+}/**'`.  I did try passing in `fileName` using encoded slashes as a workaround but this is disable in SpringBoot for security reasons.  I didn't feel comfortable sending a PR that changed that security setting as I am not very familiar with Java/Spring, but that could be an option.  Alternatively the method path could be adjusted to allow slashes by rearranging the arguments.

This was all tested and is working (with an additional change in Echo I will send) with a manually created pipeline trigger.  UI support in Deck is not implemented at the moment.

Example trigger (job is the projectId in Gitlab project):
```
 {
      "enabled": true,
      "job": "1234",
      "master": "master-from-config-file",
      "payloadConstraints": {
        "optional_contraints": "^whatever.*$"
      },
      "propertyFile": "anything",
      "type": "gitlab-ci"
}
```


README.md
```diff
@@ -164,6 +164,29 @@ that will match on output from the `jfrog rt`/`art` CLI tool.  Different regexes
 default may be configured using the `regexes` list.
 
 
+## Configuring Gitlab CI Masters
+
+In your configuration block (either in igor.yml, igor-local.yml, spinnaker.yml or spinnaker-local.yml), you can define multiple masters blocks by using the list format.
+
+To authenticate with Travis you use a "Personal Access Token" on a git user with permissions `read_api`.
```
dbyron-sf, line 171: Travis?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/GitlabCiConfig.java
```diff
@@ -40,13 +41,13 @@
 import retrofit.converter.JacksonConverter;
 
 @Configuration
-@ConditionalOnProperty("gitlab-ci.enabled")
+@ConditionalOnProperty("gitlabci.enabled")
```
dbyron-sf, line 44: This will break folks that use the old property...can we stick with the old one to prevent that?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/GitlabCiProperties.java
```diff
@@ -23,7 +23,7 @@
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.validation.annotation.Validated;
 
-@ConfigurationProperties(prefix = "gitlab-ci")
+@ConfigurationProperties(prefix = "gitlabci")
```
dbyron-sf, line 26: Same thinking here about backwards compatibility.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/GitlabCiProperties.java
```diff
@@ -50,9 +50,9 @@ public void setMasters(List<GitlabCiHost> masters) {
     @NotEmpty private String name;
     @NotEmpty private String address;
     private String privateToken;
-    private boolean limitByMembership = false;
-    private boolean limitByOwnership = true;
-    private Integer itemUpperThreshold;
+    private boolean limitByMembership = true;
```
dbyron-sf, line 53: Changing these defaults could also break folks.  I'd also love to know the reason for changing them.

linjmeyer: @dbyron-sf I don't think this code was ever fully working or finished, it doesn't work without this PR in Echo: https://github.com/spinnaker/echo/pull/1148/files and there are several closed issues related to adding Gitlab CI support that went stale.  

The default change I feel make sense though.  The defaults before required the Gitlab account that created the API token to be an owner in any project for it to appear in the API request to Gitlab.  The new defaults only require it is a member with appropriate read access.   Personally I wouldn't want to give owner access to an account just to trigger Spinnaker (that would give it full privileges to the entire project and all Gitlab features).

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/GitlabCiConfig.java
mattgogerly: Doesn't Spring convert hyphenated and camel case natively? i.e. gitlabCi === gitlab-ci. Not sure what the rational is for changing it though.

linjmeyer: Reverted to the old one and re-tested, it does work.  I think this was just a mistake on my end, learning Spring while doing this PR and I thought here were some places that were `gitlabci` and some `gitlab-ci` but looking back I don't see them on master.  I think I may have confused them with some of the `gitlab` (no CI) config?  Not sure, but either way reverted it

igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/client/logparser/PropertyParser.java
```diff
@@ -29,7 +29,7 @@
 public class PropertyParser {
 
   private static final Logger log = LoggerFactory.getLogger(PropertyParser.class);
-  private static final String MAGIC_SEARCH_STRING = "SPINNAKER_PROPERTY_";
+  private static final String MAGIC_SEARCH_STRING = ".*SPINNAKER_PROPERTY_";
```
deverton, line 32: What's the intent behind this change?

linjmeyer: Gitlab logs include the ANSI color codes, without this you can end up with garbled properties like `[32;1mMY_PROPERTY=something`, which in turn was causing invalid XML outputs.  The wildcard just ensures everything before and included SPINNAKER_PROPERTY gets removed from the key name.

dbyron-sf: Is there no way to get gitlab to stop using the color codes?  This feels like it's gonna get mis-used somehow, or be the cause of bugs, like if someone has FOO_SPINNAKER_PROPERTY_ and SPINNAKER_PROPERTY, or something like that.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/GitlabCiProperties.java
dbyron-sf: Fair enough...especially if you'll be around for the next little while to help support folks who pick up this change :)

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/GitlabCiProperties.java
```diff
@@ -50,9 +50,9 @@ public void setMasters(List<GitlabCiHost> masters) {
...
+    private boolean limitByOwnership = false;
+    private Integer itemUpperThreshold = 1000;
```
dbyron-sf, line 55: Adding a default here feels right...and I guess if we're assuming this code didn't work before, then there aren't existing users to worry about.  What happens if there are more than 1000 items though?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/gitlabci/GitlabCiBuildMonitor.java
```diff
@@ -192,8 +188,8 @@ protected void commitDelta(BuildPollingDelta delta, boolean sendEvents) {
     return pipelines.stream()
         .filter(
             pipeline ->
-                (pipeline.getFinishedAt() != null)
-                    && (pipeline.getFinishedAt().getTime() > threshold))
+                (pipeline.getCreatedAt() != null)
```
dbyron-sf, line 191: Can you help me with some terminology here.  Coming from jenkins, I create jobs and then jobs have multiple executions.  Is this talking about the creation time of the job, or an individual execution?

dbyron-sf: Either way, seems like it's worth a comment for people expecting finishedAt that the gitlab list API doesn't return it, but we still like the list API for performance reasons.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/gitlabci/client/GitlabCiClient.java
```diff
@@ -30,12 +32,32 @@
...
-      @Path("projectId") int projectId, @Query("per_page") int pageLimit);
+  List<Pipeline> getPipelineSummaries(
+      @Path("projectId") String projectId, @Query("per_page") int pageLimit);
 
   @GET("/api/v4/projects/{projectId}/pipelines/{pipelineId}")
-  Pipeline getPipeline(@Path("projectId") int projectId, @Path("pipelineId") long pipelineId);
+  Pipeline getPipeline(@Path("projectId") String projectId, @Path("pipelineId") long pipelineId);
+
+  @GET("/api/v4/projects/{projectId}/jobs/{jobId}/artifacts/{artifactPath}")
+  Response getArtifactFile(
```
dbyron-sf, line 49: Is this used anywhere?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/gitlabci/service/GitlabCiService.java
```diff
@@ -94,37 +119,116 @@ public Permissions getPermissions() {
...
+            PipelineStatus status = pipeline.getStatus();
+            if (status != PipelineStatus.running) {
+              log.error(
+                  "Unable to get GitLab build properties, pipeline '{}' in project '{}' has status {}",
+                  kv("pipeline", pipelineId),
+                  kv("project", projectId),
+                  kv("status", status));
+            }
+            // Pipeline has many jobs, jobs have many artifacts
+            // No way to list all job's artifacts so loop the jobs
```
dbyron-sf, line 174: I don't see how the code here relates to artifacts...

igor-web/src/main/groovy/com/netflix/spinnaker/igor/gitlabci/service/GitlabCiService.java
```diff
@@ -94,37 +119,116 @@ public Permissions getPermissions() {
...
+            }
+            SpinnakerException ex = new SpinnakerException(e);
+            ex.setRetryable(false);
+            throw ex;
+          } catch (IOException e) {
+            log.error("Error while parsing GitLab CI log to build properties", e);
+            return properties;
+          }
+        },
+        5,
```
dbyron-sf, line 202: Make these configurable?  Or at least add comments explaining what the params are, and how you chose these values?

igor-web/src/test/groovy/com/netflix/spinnaker/igor/gitlabci/GitlabCiBuildMonitorSpec.groovy
```diff
@@ -63,15 +63,15 @@ class GitlabCiBuildMonitorSpec extends Specification {
 
     @Ignore("Spock 1.3: The assertions on echoService never worked, but now 1.3 is smart enough to fail on that")
     @Unroll
-    def "send 2 events for a new build and store them in cache"() {
+    def "send 1 events for a new build and store them in cache"() {
```
dbyron-sf, line 66:
```suggestion
    def "send 1 event for a new build and store it in cache"() {
```

igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/client/logparser/PropertyParser.java
linjmeyer: There isn't any way I can see to change Gitlab's output but I did dig into this more and I think I have a better fix.  It's actually kind of bugged in igor/master at the moment, and that was the root cause of my issues.  I have a PR on my fork I can merge if you prefer that approach over this one [here](https://github.com/linjmeyer/igor/pull/2).  The Gitlab logs include the bash commands and their output (plus color codes) and the current setup is matching any line with the key, not just if its the start of the line only.  I fixed that (which fixes my issue) and also ensures `FOO_SPINNAKER_PROPERTY_ ` can't be used.  Check out the PR and let me know, thanks!

igor-web/src/main/groovy/com/netflix/spinnaker/igor/gitlabci/GitlabCiBuildMonitor.java
linjmeyer: In Gitlab a project has pipelines which are made up of various steps, those steps are called jobs.  Its a really weird setup to be honest, a lot of the APIs are also at the job level so you make multiple calls just to get basic info on a single pipeline run.  I can add some comments.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/GitlabCiProperties.java
linjmeyer: Same as Jenkins and others (this value gets returned in the overridden method `getPartitionUpperThreshold`).  The new builds gets thrown out and logged.  That lives in the super class method: `CommonPollingMonitor.internalPollSingle()` You can force a fast forward with `curl http://127.0.0.1:8088/admin/pollers/fastforward/gitlabCiBuildMonitor?partition=<name> -X POST`

igor-web/src/main/groovy/com/netflix/spinnaker/igor/gitlabci/client/GitlabCiClient.java
linjmeyer: It was but I hit the blocker in my main post, i'll try reintroduce in another PR that has artifact support

igor-web/src/main/groovy/com/netflix/spinnaker/igor/gitlabci/service/GitlabCiService.java
linjmeyer: made configurable (original value was copied from another polling monitor)

igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/client/logparser/PropertyParser.java
dbyron-sf: I do like https://github.com/linjmeyer/igor/pull/2 !  Thank you!

igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/GitlabCiProperties.java
dbyron-sf: Aaah, ok.  I remember this behavior being a little painful -- at least for docker images...If there are too many new ones, triggering basically stops working until there's manual intervention.  I guess this PR isn't responsible for fixing this...

igor-web/src/main/groovy/com/netflix/spinnaker/igor/gitlabci/client/GitlabCiClient.java
dbyron-sf: I'd say remove it from this PR if nothing is using it.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/gitlabci/service/GitlabCiService.java
dbyron-sf: Haven't seen a commit arrive with these changes...

dbyron-sf: Whoops...I see it now.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/gitlabci/service/GitlabCiService.java
```diff
@@ -94,37 +112,117 @@ public Permissions getPermissions() {
...
+            SpinnakerException ex = new SpinnakerException(e);
+            ex.setRetryable(false);
+            throw ex;
+          } catch (IOException e) {
+            log.error("Error while parsing GitLab CI log to build properties", e);
+            return properties;
+          }
+        },
+        this.hostConfig.getDefaultMaxHttpRetries(),
+        Duration.ofSeconds(2),
```
dbyron-sf, line 195: Sorry, mind sticking this duration in the config as well, and at least adding a comment that describes what argument false is for on the next line...and likely making that configurable too?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/gitlabci/client/GitlabCiClient.java
dbyron-sf: Sorry...you've already removed it.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/gitlabci/service/GitlabCiService.java
```diff
@@ -69,13 +77,23 @@ public BuildServiceProvider getBuildServiceProvider() {
...
+    }
+    return GitlabCiPipelineUtils.genericBuild(
+        pipeline, this.hostConfig.getAddress(), project.getPathWithNamespace());
   }
 
   @Override
-  public List<GenericBuild> getBuilds(String job) {
-    throw new UnsupportedOperationException();
+  public List<Pipeline> getBuilds(String job) {
+    return this.client.getPipelineSummaries(job, 25);
```
dbyron-sf, line 96: Sorry I didn't see these earlier.  How'd you choose 25?  Yet another thing that would benefit from a config knob?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/gitlabci/service/GitlabCiService.java
```diff
@@ -94,37 +112,117 @@ public Permissions getPermissions() {
   }
 
   public List<Project> getProjects() {
-    return getProjectsRec(new ArrayList<>(), 1);
+    return getProjectsRec(new ArrayList<>(), 1, 100).parallelStream()
```
dbyron-sf, line 115: Same question about the 100 here.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/gitlabci/service/GitlabCiService.java
```diff
@@ -94,37 +112,117 @@ public Permissions getPermissions() {
...
+            for (Job job : jobs) {
+              InputStream logStream = this.client.getJobLog(projectId, job.getId()).getBody().in();
+              String log = new String(logStream.readAllBytes(), StandardCharsets.UTF_8);
+              Map<String, Object> jobProperties = PropertyParser.extractPropertiesFromLog(log);
+              properties.putAll(jobProperties);
+            }
+
+            return properties;
+
+          } catch (RetrofitError e) {
```
dbyron-sf, line 178: Note to self to teach igor to use SpinnakerRetrofitErrorHandler...

igor-web/src/main/groovy/com/netflix/spinnaker/igor/gitlabci/service/GitlabCiService.java
linjmeyer: 100 is the largest page size you can request for Gitlab, so defaulting there makes the most sense for performance given the high number of API calls to get all the info Spinnaker expects from Gitlab.  I made this configurable as well (default still 100).

igor-web/src/main/groovy/com/netflix/spinnaker/igor/gitlabci/service/GitlabCiService.java
linjmeyer: I think it was just arbitrary to be honest.  Updated to use config option (default 100 for max page size) same as other pagination params.

linjmeyer:
Thanks for all the inputs, sorry it took me a month to catchup on all of it.  I've resolved all convos and requested a review again.  Let me know if you want any more changes!

mattgogerly:
@dbyron-sf are you happy this with? Was just looking at Echo PRs and saw one is dependent on this.

dbyron-sf:
> @dbyron-sf are you happy this with? Was just looking at Echo PRs and saw one is dependent on this.
> 
> It looks like all the requested changes have been made.

Yup, let's do it.  Sorry for the delay.

linjmeyer:
Thanks @dbyron-sf @mattgogerly 👍

PR #1049, opened 2022-10-20
fix(pipelineTriggers): change buildNumber type from int to string

jaredstehler:
in order to support build numbers which are not strict integers (i.e. "123.1")

https://github.com/spinnaker/spinnaker/issues/6743

### DIFF
--- a/igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.java
+++ b/igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.java
@@ -178,11 +178,11 @@ public class TravisBuildMonitor
 
   private Stream<? extends BuildDelta> createBuildDelta(
       String master, TravisService travisService, V3Build v3Build) {
-    int lastBuild =
+    String lastBuild =
         buildCache.getLastBuild(master, v3Build.branchedRepoSlug(), v3Build.getState().isRunning());
     return Stream.of(v3Build)
         .filter(build -> !build.spinnakerTriggered())
-        .filter(build -> build.getNumber() > lastBuild)
+        .filter(build -> build.getNumber().compareTo(lastBuild) > 0)
         .map(
             build ->
                 BuildDelta.builder()
@@ -212,9 +212,12 @@ public class TravisBuildMonitor
         if (!travisService.isLogReady(build)) {
           break;
         }
-        if (build.getNumber()
-            > buildCache.getLastBuild(
-                master, build.getRepository().getSlug(), build.getState().isRunning())) {
+        if (build
+                .getNumber()
+                .compareTo(
+                    buildCache.getLastBuild(
+                        master, build.getRepository().getSlug(), build.getState().isRunning()))
+            > 0) {
           buildCache.setLastBuild(
               master,
               build.getRepository().getSlug(),
@@ -226,7 +229,7 @@ public class TravisBuildMonitor
           }
         }
 
-        if (build.getNumber() > item.previousBuildNum) {
+        if (build.getNumber().compareTo(item.previousBuildNum) > 0) {
           buildCache.setLastBuild(
               master,
               build.branchedRepoSlug(),
@@ -311,7 +314,7 @@ public class TravisBuildMonitor
     private final V3Build build;
     private final GenericBuild genericBuild;
     private final String travisBaseUrl;
-    private final int currentBuildNum;
-    private final int previousBuildNum;
+    private final String currentBuildNum;
+    private final String previousBuildNum;
   }
 }
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/gitlabci/service/GitlabCiPipelineUtils.java
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/gitlabci/service/GitlabCiPipelineUtils.java
@@ -25,7 +25,7 @@ public class GitlabCiPipelineUtils {
     String friendlyName = String.format("%s (%s)", pipeline.getRef(), pipeline.getId());
     GenericBuild genericBuild = new GenericBuild();
     genericBuild.setBuilding(GitlabCiResultConverter.running(pipeline.getStatus()));
-    genericBuild.setNumber(pipeline.getId());
+    genericBuild.setNumber(pipeline.getId() + "");
     genericBuild.setResult(
         GitlabCiResultConverter.getResultFromGitlabCiState(pipeline.getStatus()));
     genericBuild.setName(friendlyName);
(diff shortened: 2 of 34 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/gitlabci/service/GitlabCiPipelineUtils.java
```diff
@@ -25,7 +25,7 @@ public static GenericBuild genericBuild(
     String friendlyName = String.format("%s (%s)", pipeline.getRef(), pipeline.getId());
     GenericBuild genericBuild = new GenericBuild();
     genericBuild.setBuilding(GitlabCiResultConverter.running(pipeline.getStatus()));
-    genericBuild.setNumber(pipeline.getId());
+    genericBuild.setNumber(pipeline.getId() + "");
```
dbyron-sf, line 28: Maybe `String.valueOf(pipeline.getId())` is the idea is to transform an int to a string?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/gitlabci/service/GitlabCiService.java
```diff
@@ -77,13 +77,13 @@ public List<GenericGitRevision> getGenericGitRevisions(String job, GenericBuild
...
   @Override
-  public GenericBuild getGenericBuild(String projectId, int pipelineId) {
+  public GenericBuild getGenericBuild(String projectId, String pipelineId) {
     Project project = client.getProject(projectId);
     if (project == null) {
       log.error("Could not find Gitlab CI Project with projectId={}", projectId);
       return null;
     }
-    Pipeline pipeline = client.getPipeline(projectId, pipelineId);
+    Pipeline pipeline = client.getPipeline(projectId, Long.valueOf(pipelineId));
```
dbyron-sf, line 86: Won't there be a struggle if pipelineId isn't numeric here?

igor-web/src/test/groovy/com/netflix/spinnaker/igor/gitlabci/GitlabCiBuildMonitorSpec.groovy
```diff
@@ -105,21 +105,21 @@ class GitlabCiBuildMonitorSpec extends Specification {
         service.getProjects() >> [project]
         service.getPipelines(project, 5) >> [pipeline]
         buildCache.getJobNames(MASTER) >> jobsInCache
-        buildCache.getLastBuild(MASTER, "999", true) >> lastBuildNr
+        buildCache.getLastBuild(MASTER, "999", false) >> lastBuildNr
```
dbyron-sf, line 108: Can you explain the change here?

igor-web/src/main/groovy/com/netflix/spinnaker/igor/gitlabci/service/GitlabCiService.java
jaredstehler: agreed; assumption is that gitlab only has numeric pipelineId. we could add a precondition assertion to that effect, perhaps?

igor-web/src/test/groovy/com/netflix/spinnaker/igor/gitlabci/GitlabCiBuildMonitorSpec.groovy
jaredstehler: not in the sense that I could tell how it was working previously, but by setting `true` the code was avoiding the mock stubs set up for this test.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/gitlabci/service/GitlabCiPipelineUtils.java
jaredstehler: agreed, change made.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/gitlabci/service/GitlabCiService.java
dbyron-sf: Long.valueOf throws an exception...but I bet a custom check/log message would be helpful.  Are there some gitlab docs that talk about this though?  If it turns out pipelineIds can be non-numeric, this'll be a behavior change, won't it?

igor-web/src/test/groovy/com/netflix/spinnaker/igor/gitlabci/GitlabCiBuildMonitorSpec.groovy
dbyron-sf: Aaah, ok.  I see how that works when lastBuildNr is 0, but what about when it's 100?

Would we get verification of what's happening to move this to the `then:` section and make it
```suggestion
        1 * buildCache.getLastBuild(MASTER, "999", false) >> lastBuildNr
```
?

igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.java
```diff
@@ -212,9 +212,12 @@ private void processBuild(
         if (!travisService.isLogReady(build)) {
           break;
         }
-        if (build.getNumber()
-            > buildCache.getLastBuild(
-                master, build.getRepository().getSlug(), build.getState().isRunning())) {
+        if (build
+                .getNumber()
+                .compareTo(
```
jasonmcintosh, line 217: Just a minor concern... since this is looking for making sure there aren't new builds to trigger, the compareTo with this being string vs. int now may not behave exactly the way intended.  I've not checked everyplace on this PR yet, was reading through/glancing through and type conversions where this happens... since it's a string comparison we're going to see very different results...

jasonmcintosh:
Note, the build number ordering MAY not matter, but at one point, the igor polling system did a comparison to look for any build numbers after a given pointer.  I don't think it kept ALL history, but maybe i'm wrong and it keeps all previously triggered builds?

jaredstehler:
> new Build(number: '2', timestamp: stamp1, building: false, result: 'SUCCESS'), new Build(number: '11', timestamp: stamp1, building: false, result: 'SUCCESS'), new Build(number: '22', timestamp: stamp2, building: false, result: 'SUCCESS'),
> 
> that we know that the build 11 and 22 are considered AFTER build 2... since it's now string vs. int...

should we switch to order by timestamp? or convert to int in this specific case where we know the underlying system is numeric?

jasonmcintosh:
I'd absolutely order by timestamps as a rule vs. build numbers!  As long as we can get relative timestamps.  There's a LOT in this PR, so apologies i've not dug NEAR enough to make sure not missed anything, and whether my concerns are TRULY relevant.

PR #1053, opened 2022-11-16
fix(monitor): rename parameter to let spring know what bean inject

edgarulg:
When there is more than one SpringAutoConfiguration class in the Spring context that defines a `TaskScheduler` bean, Igor fails on startup because Spring doesn't know which bean to use.

Example error message:

`***************************
APPLICATION FAILED TO START
***************************

Description:

Parameter 10 of constructor in com.netflix.spinnaker.igor.jenkins.JenkinsBuildMonitor required a single bean, but 2 were found:
	- taskScheduler: defined by method 'taskScheduler' in class path resource [org/springframework/boot/autoconfigure/task/TaskSchedulingAutoConfiguration.class]
	- vaultThreadPoolTaskScheduler: defined in null`
	
To avoid this problem we can give spring a clue on what bean has priority by include the beanName as the parameter name. In this way, we ensure that the bean injected in all Monitor classes would be the `TaskScheduler` from `TaskSchedulingAutoConfiguration`

### DIFF
--- a/igor-monitor-artifactory/src/main/java/com/netflix/spinnaker/igor/artifactory/ArtifactoryBuildMonitor.java
+++ b/igor-monitor-artifactory/src/main/java/com/netflix/spinnaker/igor/artifactory/ArtifactoryBuildMonitor.java
@@ -76,14 +76,14 @@ public class ArtifactoryBuildMonitor
       Optional<EchoService> echoService,
       ArtifactoryCache cache,
       ArtifactoryProperties artifactoryProperties,
-      TaskScheduler scheduler) {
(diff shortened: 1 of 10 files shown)


link108 (approved):

ovidiupopa07 (approved):
LGTM. It is a personal preference to use `@Qualifier` to fix such conflicts. 
At least for me, it's clearer that there is a Bean conflict and we injected a specific one instead of the `parameter name` resolution.

edgarulg:
I also like use `@Qualifier` but I just decided to change the parameter name to deliver the less changes as posible.

PR #1075, opened 2023-01-19
fix(travis): Enable legacy log fetching

jervi:
We've had lots of trouble with the `log_complete` feature Travis added a while back on our request. It was supposed to solve our issues with figuring out when logs are _actually_ available (which can be up to several minutes after the build is complete) - but basically it can't be trusted. We've seen cases where it turns to `true` even before the _build_ was completed (how can the logs be done then? 😆), and also in the opposite case, it can return `false` for a long time even though the logs _are_ in fact complete. So it's better to just fall back to the old way of checking log completion (**with some improvements***). We've been running this internally in Schibsted for several months, and have not seen any issues with log fetching during that time - a huge improvement from earlier!

\* The log complete check will now check that the log output contains one of two phrases that all Travis builds should always contain - either `Done. Your build exited with ` or `travis_terminate`.

### DIFF
--- a/igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/client/model/v3/V3Log.java
+++ b/igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/client/model/v3/V3Log.java
@@ -62,7 +62,12 @@ public class V3Log {
     }
     int numberOfParts = logParts.size() - 1;
     V3LogPart lastLogPart = logParts.get(numberOfParts);
-    return numberOfParts == lastLogPart.number && lastLogPart.isFinal();
+    String logContent = getContent();
+    return numberOfParts == lastLogPart.number
+        && lastLogPart.isFinal()
+        && (logContent != null
+            && (logContent.contains("Done. Your build exited with ")
+                || logContent.contains("travis_terminate")));
   }
 
   @JsonIgnoreProperties(ignoreUnknown = true)
--- a/igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/config/TravisConfig.java
+++ b/igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/config/TravisConfig.java
@@ -23,6 +23,7 @@ import com.netflix.spinnaker.igor.service.ArtifactDecorator;
 import com.netflix.spinnaker.igor.service.BuildServices;
 import com.netflix.spinnaker.igor.travis.TravisCache;
 import com.netflix.spinnaker.igor.travis.client.TravisClient;
+import com.netflix.spinnaker.igor.travis.client.model.v3.Root;
 import com.netflix.spinnaker.igor.travis.service.TravisService;
 import com.netflix.spinnaker.retrofit.Slf4jRetrofitLogger;
 import com.squareup.okhttp.OkHttpClient;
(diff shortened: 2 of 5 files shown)


jervi:
@spinnaker/oss-reviewers PTAL 🙂

igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/config/TravisConfig.java
```diff
@@ -83,17 +82,22 @@ public Map<String, TravisService> travisMasters(
                               igorConfigurationProperties.getClient().getTimeout(),
                               objectMapper);
 
-                      boolean useLegacyLogFetching = false;
```
dbyron-sf, line 86: For the record, it looks like this came from https://github.com/spinnaker/igor/pull/449.

igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/client/model/v3/V3Log.java
```diff
@@ -62,7 +62,12 @@ public boolean isReady() {
     }
     int numberOfParts = logParts.size() - 1;
     V3LogPart lastLogPart = logParts.get(numberOfParts);
-    return numberOfParts == lastLogPart.number && lastLogPart.isFinal();
+    String logContent = getContent();
```
dbyron-sf, line 65: This seems like a new change...not reverting part of https://github.com/spinnaker/igor/pull/449 ...  Is this at least sort of, a separate improvement?

igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/config/TravisConfig.java
```diff
@@ -83,17 +82,22 @@ public Map<String, TravisService> travisMasters(
...
-                        Root root = client.getRoot();
-                        useLegacyLogFetching = !root.hasLogCompleteAttribute();
-                        if (useLegacyLogFetching) {
-                          log.info(
-                              "It seems Travis Enterprise is older than version 2.2.9. Will use legacy log fetching.");
-                        }
-                      } catch (Exception e) {
-                        log.warn("Could not query Travis API to check API compability", e);
-                      }
+                      // Disabling the use of the log_complete attribute for now. We suspect it can
```
dbyron-sf, line 85: Having commented-out code like this isn't my favorite.  Does a new config knob like "forceLegacyLogFetching" make sense?  Or "just" deleting the code and depending on git to keep the old stuff for us?

igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/client/model/v3/V3Log.java
jervi: Correct, this is an improvement. Travis will always end a build (failed or not) like this.

igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/config/TravisConfig.java
jervi: Yes, correct. We had high hopes for this feature 😢

igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/config/TravisConfig.java
jervi: I'm fine with deleting it actually, but then I should probably revert the rest of #449 as as well, don't you think?

dbyron-sf: I don't know enough to say.  https://github.com/spinnaker/igor/pull/449 is pretty big...and maybe there are some bits that make sense to keep?

jervi: I'll have another look and see. If it gets too complicated, I can probably just go with the force knob thing.

jervi: I've added a feature flag now 👍

dbyron-sf: Thanks!

dbyron-sf (approved):

PR #1077, opened 2023-01-26
fix(travis): Redis TTL is given in seconds, not minutes

jervi:
This fixes a rare bug where some builds will "be lost" and not trigger any pipelines. Because we effectively set the TTL to 5 minutes instead of 5 hours, Spinnaker would quickly stop tracking builds. Usually this is no problem because they will be present in the builds returned from the Travis API, but if the `numberOfJobs` property is set too low or during periods with a lot of simultaneous builds, this could cause builds (especially long running ones) to be lost.

### DIFF
--- a/igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.java
+++ b/igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.java
@@ -68,7 +68,7 @@ public class TravisBuildMonitor
   private final TravisProperties travisProperties;
   private final Optional<EchoService> echoService;
 
-  public static final int TRACKING_TTL = (int) TimeUnit.HOURS.toMinutes(5);
+  public static final int TRACKING_TTL = (int) TimeUnit.HOURS.toSeconds(5);
 
   @Autowired
   public TravisBuildMonitor(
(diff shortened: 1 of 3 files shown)


igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/TravisBuildMonitor.java
```diff
@@ -68,7 +68,7 @@
   private final TravisProperties travisProperties;
   private final Optional<EchoService> echoService;
 
-  public static final int TRACKING_TTL = (int) TimeUnit.HOURS.toMinutes(5);
+  public static final int TRACKING_TTL = (int) TimeUnit.HOURS.toSeconds(5);
```
dbyron-sf, line 71: Thank you!

Looks like this could be package private / only used in TravisBuildMonitorSpec.  You up to rename this to
```suggestion
  static final int TRACKING_TTL_SECS = (int) TimeUnit.HOURS.toSeconds(5);
```
and maybe even adding secs to the args for BuildCache.setTracking + setTtl so it's more obvious?

jervi: Done 👍

dbyron-sf (approved):

PR #1089, opened 2023-02-21
fix(travis): Don't panic if log fetching fails

jervi:
If a user has deleted the log for a job in Travis or it has expired due to old age, Spinnaker may fail to display the list of builds in the start manual execution box, even though there are other builds/jobs available. This fix will just filter these (and other) failures, and instead of failing completely, it will print an error message in the Igor log. The `catch` clause will print a log line at level "INFO" for cases where the log has expired (which can happen from time to time) and at WARN level for other (actual) errors.

image

### DIFF
--- a/igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/service/TravisService.java
+++ b/igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/service/TravisService.java
@@ -59,6 +59,7 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -428,6 +429,7 @@ public class TravisService implements BuildOperations, BuildProperties {
     return jobIds.stream().map(this::getAndCacheJobLog).allMatch(Optional::isPresent);
   }
 
+  @SuppressWarnings("rawtypes")
   private Optional<String> getAndCacheJobLog(int jobId) {
     log.debug("fetching log by jobId {}", jobId);
     String cachedLog = travisCache.getJobLog(groupKey, jobId);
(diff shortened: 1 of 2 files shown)


jervi:
@spinnaker/oss-reviewers PTAL :)

igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/service/TravisService.java
```diff
@@ -428,18 +429,41 @@ public boolean isLogReady(List<Integer> jobIds) {
...
-      travisCache.setJobLog(groupKey, jobId, v3Log.getContent());
-      return Optional.of(v3Log.getContent());
+    try {
+      V3Log v3Log = travisClient.jobLog(getAccessToken(), jobId);
+      if (v3Log != null && v3Log.isReady()) {
+        log.info("Log for jobId {} was ready, caching it", jobId);
+        travisCache.setJobLog(groupKey, jobId, v3Log.getContent());
+        return Optional.of(v3Log.getContent());
+      }
+    } catch (RetrofitError e) {
```
dbyron-sf, line 447: I'd of course like to see tests that demonstrate that this error handling (which I like) is working.  I'm not sure if there are good models to follow in igor.  Maybe [ArtifactDownloaderImplTest from rosco](https://github.com/spinnaker/rosco/blob/master/rosco-manifests/src/test/java/com/netflix/spinnaker/rosco/manifests/ArtifactDownloaderImplTest.java) is helpful?

I'm slightly hesitant to add more code that depends on RetrofitError.  It's not fair of me to ask for https://github.com/spinnaker/spinnaker/issues/5473 first, though I do want to point it out.  I do happen to know of work going on specifically for igor for that issue, and I suppose this won't make too much more work.

jervi: Thanks, David! I wasn't aware of that refactor task, but I totally agree that it is something we should do. But maybe not as a part of this PR. But I'll see if I can add a test.

jervi: Test added 👍

dbyron-sf (approved):

PR #1090, opened 2023-02-24
refactor(travisclient): make changes to travisClient to use Spinnaker custom exception instead of RetrofitError using SpinnakerRetrofitErrorHandler

Luthan95:
To make use of custom spinnaker exception instead of RetrofitError using SpinnakerRetrofitErrorHandler present in kork service, to rectify dependency of RetrofitError class

### DIFF
--- a/igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/service/TravisService.java
+++ b/igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/service/TravisService.java
@@ -50,6 +50,8 @@ import com.netflix.spinnaker.igor.travis.client.model.v3.V3Build;
 import com.netflix.spinnaker.igor.travis.client.model.v3.V3Builds;
 import com.netflix.spinnaker.igor.travis.client.model.v3.V3Job;
 import com.netflix.spinnaker.igor.travis.client.model.v3.V3Log;
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerHttpException;
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerServerException;
 import io.github.resilience4j.circuitbreaker.CircuitBreaker;
 import io.github.resilience4j.circuitbreaker.CircuitBreakerRegistry;
 import io.github.resilience4j.core.SupplierUtils;
@@ -59,7 +61,6 @@ import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Comparator;
-import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -73,7 +74,6 @@ import net.logstash.logback.argument.StructuredArguments;
 import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import retrofit.RetrofitError;
 
 public class TravisService implements BuildOperations, BuildProperties {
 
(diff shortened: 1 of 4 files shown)


igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/service/TravisService.java
```diff
@@ -444,26 +443,20 @@ private Optional<String> getAndCacheJobLog(int jobId) {
...
-            log.warn(
-                "{}: Could not get log for job id {}. Error from Travis:\n{}",
-                groupKey,
-                jobId,
-                body);
-          }
-        } catch (RuntimeException ex) {
-          log.warn("{}: Could not parse original error message from Travis", groupKey, ex);
+    } catch (SpinnakerServerException e) {
+      if (e.getResponseBody() != null) {
```
dbyron-sf, line 447: Let's call `e.getResponseBody()` once and check the resulting variable for null.

igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/service/TravisService.java
```diff
@@ -444,10 +444,10 @@ private Optional<String> getAndCacheJobLog(int jobId) {
...
         return Optional.of(v3Log.getContent());
       }
-    } catch (RetrofitError e) {
-      if (e.getBody() != null) {
+    } catch (SpinnakerServerException e) {
+      if (e instanceof SpinnakerHttpException) { // only SpinnakerHttpException has a response body
         try {
-          Map body = (Map) e.getBodyAs(HashMap.class);
+          Map<String, Object> body = ((SpinnakerHttpException) e).getResponseBody();
           if ("log_expired".equals(body.get("error_type"))) {
```
dbyron-sf, line 451: We need a null check for body, yes?

igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/service/TravisService.java
```diff
@@ -444,11 +444,11 @@ private Optional<String> getAndCacheJobLog(int jobId) {
         travisCache.setJobLog(groupKey, jobId, v3Log.getContent());
         return Optional.of(v3Log.getContent());
       }
-    } catch (RetrofitError e) {
-      if (e.getBody() != null) {
+    } catch (SpinnakerServerException e) {
+      if (e instanceof SpinnakerHttpException) { // only SpinnakerHttpException has a response body
         try {
```
dbyron-sf, line 449: Can we get rid of this try/catch block now that we're no long called e.getBodyAs?

igor-monitor-travis/src/main/java/com/netflix/spinnaker/igor/travis/service/TravisService.java
```diff
@@ -444,27 +444,21 @@ private Optional<String> getAndCacheJobLog(int jobId) {
...
-            log.warn(
-                "{}: Could not get log for job id {}. Error from Travis:\n{}",
-                groupKey,
-                jobId,
-                body);
-          }
-        } catch (RuntimeException ex) {
-          log.warn("{}: Could not parse original error message from Travis", groupKey, ex);
+    } catch (SpinnakerServerException e) {
+      if (e instanceof SpinnakerHttpException) { // only SpinnakerHttpException has a response body
```
dbyron-sf, line 448: What you've got here matches the current behavior, but...I'm tempted to clean up this code and change the behavior all at the same time.

If I'm reading the code correctly, the current behavior swallows an RetrofitError that has a null body.  That seems a little unfortunate.

If we change this to catch SpinnakerHttpException instead of SpinnakerServerException, we let the other exception types bubble up.  That feels like better behavior.  But, this PR has lived a long time, and I don't have a good way to exercise this specific functionality so let's leave this for another time.

dbyron-sf (approved):

PR #1143, opened 2023-06-27
chore(build): default containers to JRE 17

mattgogerly:
Default containers will use JRE 17. Versions using JRE 11 will also be published to allow opting out if required.


mattgogerly:
@Mergifyio update

PR #1170, opened 2023-09-18
fix(java17): remove long-deprecated import that doesn't work with JRE17

mattgogerly:
`sun.net.*` is no longer accessible [as of JDK16](https://openjdk.org/jeps/396). Thankfully `    Security.setProperty('networkaddress.cache.ttl', '0')` seems to achieve the same goal as the removed code, so I'm not sure why they were both there in the first place.

Closes https://github.com/spinnaker/spinnaker/issues/6879.


dbyron-sf (approved):

dbyron-sf:
@mergifyio backport release-1.32.x

PR #1173, opened 2023-09-20
fix(java17): resolve remaining access to private fields, run tests on JRE 17

mattgogerly:
`fastTime` and `size` are no longer accessible fields. Replace with `getTime()` and `size()` respectively. IntelliJ shows no remaining access related issues (not to say there aren't more hiding).

Additionally, run tests using JRE 17 to catch this stuff earlier.

Note: this does **not** resolve the issue reported relating to `Instant.seconds` - further PR to come in Kork for that one.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -25,7 +25,9 @@ jobs:
         uses: docker/setup-buildx-action@v2
       - uses: actions/setup-java@v3
         with:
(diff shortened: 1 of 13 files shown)


mattgogerly:
@Mergifyio update

jasonmcintosh (approved):

PR #1174, opened 2023-09-20
fix(java17): add Jackson converter to RestAdapters to avoid Gson

mattgogerly:
Fixes https://github.com/spinnaker/spinnaker/issues/6881

If no converter is set for a RestAdapter it defaults to Gson, which uses reflection to set fields. JRE 17 says no bueno to that on classes like `Instant` that are now making fields inaccessible.


mattgogerly:
@Mergifyio update

jasonmcintosh (approved):

mattgogerly:
@Mergifyio backport release-1.32.x

PR #1130, opened 2023-05-22
refactor(JenkinsClient): make changes to jenkinsclient to use Spinnak…

Luthan95:
…er custom exception instead of RetrofitError using SpinnakerRetrofitErrorHandler

Make Jekinsclient use of spinnakerCustomexception by setting SpinnakerRetrofitErrorHandler as ErrorHandler, so avoid direct use of RetrofitError

PR #1187, opened 2023-10-26
fix(jenkins): replacing retrofit error 

xibz:
igor relies on the SpinnakerRetrofitHandler and it looks like this one
got missed, which would cause the `getBuilds` method to fail if an
exception occurred rather than happily carrying on.

### DIFF
--- a/igor-web/igor-web.gradle
+++ b/igor-web/igor-web.gradle
@@ -43,6 +43,7 @@ dependencies {
     implementation "org.codehaus.groovy:groovy"
 
     implementation "com.fasterxml.jackson.core:jackson-annotations"
+    implementation 'com.fasterxml.jackson.module:jackson-module-jaxb-annotations:2.13.1'
     implementation "com.fasterxml.jackson.core:jackson-core"
     implementation "com.fasterxml.jackson.core:jackson-databind"
     implementation "com.fasterxml.jackson.dataformat:jackson-dataformat-xml"
--- a/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
+++ b/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
@@ -35,6 +35,7 @@ import com.netflix.spinnaker.igor.polling.PollingDelta
 import com.netflix.spinnaker.igor.service.BuildServices
 import com.netflix.spinnaker.kork.discovery.DiscoveryStatusListener
 import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerServerException
 import com.netflix.spinnaker.security.AuthenticatedRequest
 import groovy.time.TimeCategory
 import org.springframework.beans.factory.annotation.Autowired
@@ -42,7 +43,6 @@ import org.springframework.beans.factory.annotation.Value
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.scheduling.TaskScheduler
 import org.springframework.stereotype.Service
-import retrofit.RetrofitError
 import java.util.stream.Collectors
 
 import static net.logstash.logback.argument.StructuredArguments.kv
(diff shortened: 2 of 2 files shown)


igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
```diff
@@ -35,6 +35,9 @@ import com.netflix.spinnaker.igor.polling.PollingDelta
 import com.netflix.spinnaker.igor.service.BuildServices
 import com.netflix.spinnaker.kork.discovery.DiscoveryStatusListener
 import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService
+import com.netflix.spinnaker.kork.exceptions.SpinnakerException
+import com.netflix.spinnaker.kork.retrofit.exceptions.SpinnakerServerException
+import com.netflix.spinnaker.kork.retrofit.exceptions.UpstreamBadRequest
```
dbyron-sf, line 40: Doesn't look like this is used here.

igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy
```diff
@@ -157,9 +160,11 @@ class JenkinsBuildMonitor extends CommonPollingMonitor<JobDelta, JobPollingDelta
 
         } catch (e) {
             log.error("Error processing builds for [{}:{}]", kv("master", master), kv("job", job.name), e)
-            if (e.cause instanceof RetrofitError) {
-                def re = (RetrofitError) e.cause
-                log.error("Error communicating with jenkins for [{}:{}]: {}", kv("master", master), kv("job", job.name), kv("url", re.url), re)
+            if (e instanceof SpinnakerServerException) {
+                def re =  e.getRetrofitError();
```
dbyron-sf, line 164: I'd really like to leave any use of retrofit1 classes behind here, so we can upgrade to retrofit2, remove retrofit1, and remove exposure to some CVEs.  At the moment, [SpinnakerServerException](https://github.com/spinnaker/kork/blob/8433e316bdcc7611c8bd64e5611c84ca7fedebb7/kork-retrofit/src/main/java/com/netflix/spinnaker/kork/retrofit/exceptions/SpinnakerServerException.java) doesn't have a url attribute, but it could, and then `e.url` would work here.

dbyron-sf:
From what I can tell, this isn't really tied to https://github.com/spinnaker/kork/pull/1108.  It's the [getBuilds](https://github.com/spinnaker/igor/blob/8ff1cc3099cd76a2d3fa442890c349c255a3ec44/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy#L138) call inside the try/catch block that could throw a Spinnaker*Exception, via [jenkinsService.getBuilds](https://github.com/spinnaker/igor/blob/8ff1cc3099cd76a2d3fa442890c349c255a3ec44/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/JenkinsBuildMonitor.groovy#L170-L174).  The jenkinsClient inside [JenkinsService](https://github.com/spinnaker/igor/blob/8ff1cc3099cd76a2d3fa442890c349c255a3ec44/igor-web/src/main/groovy/com/netflix/spinnaker/igor/jenkins/service/JenkinsService.java#L165) comes from [JenkinsConfig](https://github.com/spinnaker/igor/blob/8ff1cc3099cd76a2d3fa442890c349c255a3ec44/igor-web/src/main/groovy/com/netflix/spinnaker/igor/config/JenkinsConfig.groovy#L198), which already (as ofhttps://github.com/spinnaker/igor/pull/1130) uses SpinnakerRetrofitErrorHandler.

igor-web/igor-web.gradle
```diff
@@ -43,6 +43,7 @@ dependencies {
     implementation "org.codehaus.groovy:groovy"
 
     implementation "com.fasterxml.jackson.core:jackson-annotations"
+    implementation 'com.fasterxml.jackson.module:jackson-module-jaxb-annotations:2.13.1'
```
dbyron-sf, line 46: Do we still need this?

dbyron-sf:
included in https://github.com/spinnaker/igor/pull/1188

PR #1204, opened 2023-12-14
feat(jenkins): Enable Jenkins job triggers for jobs in sub-folders

ciurescuraul:
### Description
When defining a Jenkins job in a sub-folder, the path contains slashes. Consequently, the Spring framework is not capable of accurately matching the request.

### Problem
The trigger request fails because it is attempting to initiate a job that is not found. 

### Solution
A feature flag has been added. This flag influences if we should call the existing endpoint (which accepts the job name as a path variable) or the new one (which accepts the job name as a query parameter).

### Before
image

### After
image

### Related PR's
Orca: https://github.com/spinnaker/orca/pull/4618
Echo: https://github.com/spinnaker/echo/pull/1373


ovidiupopa07 (approved):
LGTM

jasonmcintosh (approved):
Looks good (only question from others is encoding, but think this'll work pretty well)

ovidiupopa07:
@Mergifyio backport release-1.28.x release-1.30.x release-1.31.x release-1.32.x release-1.33.x

PR #1218, opened 2024-01-26
Telemetry: fix polling gauges always reporting NaN on Prometheus and timers

massimo-pacher-tw:
Fixes https://github.com/armory-plugins/armory-observability-plugin/issues/34. 
Long overdue, I know, as we have been running with it for more than a year now and forgot to upstream it 😓  

# Context

Relevant polling gauges are currently always **NaN**: they represent new items (e.g images cached) or above the limit, requiring manual intervention (via admin endpoint).

The lack of visibility over these metrics (just visible in logs) might be a big issues, given blocked caching might have consequences like engineers not be able to find they releases or triggers not firing.

The PR fixes this using the spectator library, trying to achieve a similar outcome we use internally at Wise in our production services with Micrometer, i.e. registering some variables with strong references and letting Prometheus scrapes observe the latest value.

In order to do that, we introduced a separate instrumentation class, which eventually could contain logging too, as per [Domain Probes ](https://martinfowler.com/articles/domain-oriented-observability.html).

Testing tried to simulate multiple runs and simple concurrency scenarios.

**P.S. I wrote this long time ago and cleaned up the internal PR, so there might be some caveats based on the internal use I might have missed but it does fix the issue reported.**

### DIFF
--- a/igor-core/src/main/java/com/netflix/spinnaker/igor/polling/CommonPollingMonitor.java
+++ b/igor-core/src/main/java/com/netflix/spinnaker/igor/polling/CommonPollingMonitor.java
@@ -24,13 +24,17 @@ import com.netflix.spinnaker.kork.discovery.DiscoveryStatusListener;
 import com.netflix.spinnaker.kork.discovery.RemoteStatusChangedEvent;
 import com.netflix.spinnaker.kork.dynamicconfig.DynamicConfigService;
 import java.time.Duration;
+import java.util.Map;
 import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import javax.annotation.Nullable;
 import javax.annotation.PreDestroy;
 import net.logstash.logback.argument.StructuredArguments;
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.scheduling.TaskScheduler;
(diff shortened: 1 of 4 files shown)


igor-core/src/main/java/com/netflix/spinnaker/igor/polling/CommonPollingMonitor.java
```diff
@@ -163,7 +163,17 @@ protected String getLockName(String name, String partition) {
   }
 
   protected void internalPollSingle(PollContext ctx) {
-    String monitorName = getClass().getSimpleName();
+    String monitorName =
+        !StringUtils.isBlank(this.getName()) ? this.getName() : getClass().getSimpleName();
```
massimo-pacher-tw, line 167: I think this is was a bug, cause overridden function was totally ignored in metrics

igor-core/src/main/java/com/netflix/spinnaker/igor/polling/CommonPollingMonitor.java
```diff
@@ -67,12 +71,9 @@ public CommonPollingMonitor(
     this.discoveryStatusListener = discoveryStatusListener;
     this.lockService = lockService;
     this.scheduler = scheduler;
+    this.instrumentation = new CommonPollingMonitorInstrumentation(registry);
```
massimo-pacher-tw, line 74: single instance per monitor, I know it's not needed, but given we just run a few, it's not a biggy, compared to fix all the tests and change the constructor of every subclass.
The key bit is the registry, which can be passed in for testing

igor-core/src/main/java/com/netflix/spinnaker/igor/polling/CommonPollingMonitorInstrumentation.java
```diff
@@ -0,0 +1,99 @@
...
+  public CommonPollingMonitorInstrumentation(Registry registry) {
+    this.registry = registry;
+    itemsCachedId = registry.createId("pollingMonitor.newItems");
+    itemsOverThresholdId = registry.createId("pollingMonitor.itemsOverThreshold");
+    pollCycleFailedId = registry.createId("pollingMonitor.failed");
+    pollCycleTimingId = registry.createId("pollingMonitor.pollTiming");
+  }
+
+  public void trackItemsCached(AtomicInteger numberOfItems, String monitor, String partition) {
+    Gauge gauge =
```
jasonmcintosh, line 42: Minor... but could move the registry.get into the if check bypassing the need for a cast

igor-core/src/main/java/com/netflix/spinnaker/igor/polling/CommonPollingMonitorInstrumentation.java
```diff
@@ -0,0 +1,99 @@
...
+    if (gauge == null) {
+      PolledMeter.using(registry)
+          .withId(itemsCachedId.withTags("monitor", monitor, "partition", partition))
+          .monitorValue(numberOfItems);
+    }
+  }
+
+  public void trackItemsOverThreshold(
+      AtomicInteger numberOfItems, String monitor, String partition) {
+    Gauge gauge =
```
jasonmcintosh, line 64: same as above, minor thing just prefer inline checks on some of this :) PERSONAL thing vs. "required"

jasonmcintosh (approved):
Looks like this covers the sitaution :) Thank you!

PR #1256, opened 2024-05-08
fix(travis): Properly return API triggered builds

jervi:
When populating the build dropdown for manual executions, manually triggered builds are missing for Travis Enterprise version 3.0.54 and newer. Before these builds had the same `event_type=push` as other builds, but now they have `event_type=api`. Adding `api` to the filtering fixes the issue.


dbyron-sf:
What about older versions of Travis Enterprise?

jervi:
> What about older versions of Travis Enterprise?

Luckily this filter accepts anything, so it will just silently ignore any unknown event types.

dbyron-sf (approved):

PR #1258, opened 2024-05-31
refactor(artifact): moved artifacts (groovy) unit tests to artifacts (java)

aman-agrawal:
- moved ArtifactExtractorSpec groovy tests to ArtifactExtractorTest java tests
- Newly added TCs are passing in ArtifactExtractorTest
- deleted ArtifactExtractorSpec
- changed parseCustomFormat() from private to package-private (default) to access it in testParseCustomFormatCorrectlyParsesBooleansAndStrings() & added `@VisibleForTesting`

--
- moved ArtifactServiceSpec groovy tests to ArtifactServiceTest java tests
- Newly added TCs are passing in ArtifactServiceTest
- deleted ArtifactServiceSpec
- ArtifactServiceSpec method `"service finds artifact versions"() `called `getArtifactVersions(String, String, List<String>)`
To call same method from ArtifactServiceTest and to resolve the ambiguity issue between the overloaded methods, typecasted null with type `List<String>`.


igor-web/src/main/groovy/com/netflix/spinnaker/igor/artifacts/ArtifactExtractor.java
```diff
@@ -91,7 +91,7 @@ private JinjaTemplate getTemplateFromProperty(GenericBuild build) {
     return jinjaTemplateService.getTemplate(messageFormat, templateType);
   }
 
-  private boolean parseCustomFormat(Object customFormat) {
+  protected boolean parseCustomFormat(Object customFormat) {
```
dbyron-sf, line 94: If this is only for testing, please add `@VisibleForTesting`...except I don't see anything inheriting from ArtifactExtractor so I'm not sure this is necessary.

dbyron-sf: Maybe package-private?

aman-agrawal: Done

igor-web/src/main/java/com/netflix/spinnaker/igor/artifacts/ArtifactService.java
```diff
@@ -37,7 +37,8 @@ public interface ArtifactService {
   default List<String> getArtifactVersions(
       @Nonnull String type, @Nonnull String name, String releaseStatus) {
     List<String> releaseStatuses = new ArrayList<>();
-    releaseStatuses.add(releaseStatus);
+    if (releaseStatus == null) releaseStatuses = null;
```
dbyron-sf, line 40: This is more than a refactor.  Can we leave the logic changes for a subsequent PR please?  It also seems like a strange line of code to add...does it do anything?

aman-agrawal: **Change reverted** as this method is not called from unit test.
ArtifactServiceSpec method `"service finds artifact versions"()` called `getArtifactVersions(String, String, List<String>)`
To call same method from ArtifactServiceTest and to resolve the ambiguity issue between the overloaded methods, typecasted null with type `List<String>` in `serviceFindsArtifactVersions()`

dbyron-sf (approved):

PR #1274, opened 2024-09-05
refactor(dependency): replace groovy coordinates during upgrade of groovy 4.x

j-sandy:
Replacing the groovy coordinates from `org.codehaus.groovy` to `org.apache.groovy` supported by groovy 4.x and above versions. Excluding the transitive groovy dependency from `org.jfrog.artifactory.client:artifactory-java-client-services:2.9.2`, and explictly adding the new coordinates of groovy to igor-monitor-artifactory module.


dbyron-sf:
included in https://github.com/spinnaker/igor/pull/1275

PR #1283, opened 2024-11-07
chore(build): target Java 17 bytecode

mattgogerly:
Update the build to target Java 17 bytecode when building.

cc @dbyron-sf

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -2,7 +2,7 @@ fiatVersion=1.50.0
 korkVersion=7.243.0
 org.gradle.parallel=true
 spinnakerGradleVersion=8.32.1
-targetJava17=false
+targetJava17=true
 
 # To enable a composite reference to a project, set the
 #  project property `'<projectName>Composite=true'`.


gradle.properties
```diff
@@ -2,7 +2,7 @@ fiatVersion=1.50.0
 korkVersion=7.243.0
 org.gradle.parallel=true
 spinnakerGradleVersion=8.32.1
-targetJava17=false
+targetJava17=true
```
dbyron-sf, line 5: Is it also time to remove the java 11 docker images?

dbyron-sf: Plus also, do we still need `-PenableCrossCompilerPlugin=true` in the gradle invocations?

mattgogerly: I reckon we go live with this small change in Igor, and if we're happy we come back and purge Java 11 from existence a few weeks after release.

Then for other repos just switch/purge.

dbyron-sf: OK, let's do it. This'll make it into 1.36....

dbyron-sf (approved):

PR #1304, opened 2025-01-27
fix(jenkins): Use Long values rather than Ints for buildNumbers and Queue numbers

michaelepperson:
This PR is to resolve [#7002](https://github.com/spinnaker/spinnaker/issues/7002).

Cloudbees has started using Long for HA enabled Queues. To resolve this, Starting from the getQueue code, I have change the expected type from Int to Long. I followed the trail through until all changes needed were made.

### DIFF
--- a/igor-core/src/main/java/com/netflix/spinnaker/igor/build/BuildCache.java
+++ b/igor-core/src/main/java/com/netflix/spinnaker/igor/build/BuildCache.java
@@ -71,14 +71,14 @@ public class BuildCache {
     return results;
   }
 
-  public int getLastBuild(String master, String job, boolean running) {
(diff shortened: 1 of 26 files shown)


jasonmcintosh (approved):
This should cover all the areas for coms.  I don't know that we needed travis, but since this is a generic API change, this should help across the board to support longer build numbers.

jasonmcintosh:
ALSO:  There will be needed some Orca work to handle longer times as well.  E.g.:
https://github.com/spinnaker/orca/blob/33097a3ae6de11926cdd0187f20d81cd2251e368/orca-igor/src/main/java/com/netflix/spinnaker/orca/igor/IgorService.java#L42

Separate PR for that should be incoming.

